PR #3, opened 2023-10-28
feat: add exec committee images

BastiDood:
Just to put something onto the screen when the user visits the "People" page, I have added the Executive Committee. In the future, we will add dedicated Committee pages that will allow our members to post their names and their social media links.


Anjellyrika (approved):
LGTM! 🌟

PR #5, opened 2023-12-11
feat: create carousel and its card as components

Airiseru:
Create carousel cards, with placeholder content, and placed in the carousel component

### DIFF
--- a/src/lib/components/CarouselCard.svelte
+++ b/src/lib/components/CarouselCard.svelte
@@ -1,0 +1,4 @@
+<!-- Carousel Card with filler content -->
+<div class="font-dm flex-initial w-56">
+    <p>Lorem ipsum dolor sit amet consectetur, adipisicing elit. Similique dignissimos maxime eligendi eaque non, aperiam nihil iste, odio autem eum fugiat veritatis ipsum inventore atque nisi est, reprehenderit consequatur ex?</p>
+</div>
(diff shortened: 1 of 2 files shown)


src/lib/components/CarouselCard.svelte
```diff
@@ -0,0 +1,8 @@
+<!-- Carousel Card with filler content -->
+<div class="w-56 flex-initial rounded-full bg-blue p-5 text-center font-dm">
+    <!-- Sample Text -->
+    <p class="m-0">Lorem ipsum dolor sit amet consectetur</p>
+
+    <!-- Sample Logo -->
+    <!-- <img src="https://upload.wikimedia.org/wikipedia/commons/thumb/e/e1/Logo_of_YouTube_%282015-2017%29.svg/2560px-Logo_of_YouTube_%282015-2017%29.svg.png" alt="" class="m-auto w-40"> -->
```
BastiDood, line 7: Normally, since we have version control, we can actually remove lines of code that are essentially commented out. There's no reason to keep them because that's what `git checkout` is for.
```suggestion
    <!-- Sample Logo -->
```

BastiDood (changes requested):
Everything looks great here! Just one nitpick before we can merge things here.

BastiDood (approved):
LGTM! 🚀

PR #6, opened 2023-12-12
feat: add content container component

KindaExists:
This PR adds a content container component.


BastiDood (approved):
Awesome work! Huge bonus points for even adding media queries into the mix. In future iterations of this component, we should definitely consider using [slots](https://learn.svelte.dev/tutorial/slots) instead. For now, this work is already beyond my expectations. Great job! 🚀

PR #7, opened 2023-12-12
feat: add buttons and footer

rue-22:
This PR adds the footer and button components for the website.


src/lib/components/Footers.svelte
```diff
@@ -1,8 +1,8 @@
 <script>
-    import CFooter from "$lib/components/CFooter.svelte";
-    import SFooter from "$lib/components/SFooter.svelte"
+    import CFooter from '$lib/components/CFooter.svelte';
+    import SFooter from '$lib/components/SFooter.svelte';
```
BastiDood, line 3: Just a nitpick, but it would be great if we could use a more descriptive import name and file name for the footer. 👌

src/lib/components/SFooter.svelte
```diff
@@ -1,20 +1,23 @@
...
-        <br>
+        <br />
+        <a
+            class="underline"
+            href="https://www.linkedin.com/company/up-center-for-student-innovations/">LinkedIn</a
+        >
+        <br />
         <a class="underline" href="https://github.com/up-csi">GitHub</a>
-        <p class="text-black font-inter mt-4">Site built with SvelteKit</p>
+        <p class="mt-4 font-inter text-black">Site built with SvelteKit</p>
```
BastiDood, line 21: In hindsight, we probably don't need to disclose this.
```suggestion
```

src/lib/components/Buttons.svelte
```diff
@@ -0,0 +1,7 @@
+<script>
+    import Button1 from '$lib/components/Button1.svelte';
+    import Button2 from '$lib/components/Button2.svelte';
```
BastiDood, line 3: We could use a more descriptive file name and import name for this. Perhaps `CsiBlueButton.svelte` and `BlackButton.svelte` would be nice?

src/lib/components/Button1.svelte
```diff
@@ -0,0 +1,4 @@
+<!-- Dark Button -->
```
BastiDood, line 1: No need for this comment here as it's already self-explanatory.
```suggestion
```

src/lib/components/Button2.svelte
```diff
@@ -0,0 +1,4 @@
+<!-- Blue Button -->
```
BastiDood, line 1: No need for this comment here as it's already self-explanatory.
```suggestion
```

src/lib/components/SFooter.svelte
```diff
@@ -0,0 +1,23 @@
+<div class="flex h-auto w-auto justify-between font-inter">
+    <div class="text-left">
+        <p class="text-lg">UP CSI is a service-</p>
+        <p class="text-lg">oriented organization</p>
+        <p class="text-lg">for student learning</p>
```
BastiDood, line 5: Maybe we could instead rely on `max-width` here rather than manually writing the line breaks ourselves.

BastiDood (changes requested):

BastiDood (changes requested):
Seems like some of the old files remained during the `git merge`. I've set the conversations as unresolved until we remove the old ones. 👍

src/lib/components/Buttons.svelte
BastiDood, line 1: You know, I've been thinking about this file. Maybe if it's just gonna be two components, we might as well use the inner components directly in the actual page.

src/lib/components/Footers.svelte
```diff
@@ -0,0 +1,8 @@
+<script>
+    import CenterFooter from '$lib/components/CenterFooter.svelte';
+    import SidesFooter from '$lib/components/SidesFooter.svelte';
+</script>
+
+<SidesFooter />
+<br />
+<CenterFooter />
```
BastiDood, line 8: I think it's worth considering that we use flexbox here rather than an explicit `<br />`. And also, we should probably just inline these two footers rather than putting them as dedicated components. That would make the flexbox styling easier.

BastiDood (changes requested):
Thanks, just for a few more notes and we're ready to have this merged.

src/lib/components/SidesFooter.svelte
BastiDood, line 1: Same goes for this file, by the way. It would be great to prefer flexbox or grid instead of the manual `<br />` breaks.

BastiDood (changes requested):

src/lib/components/SidesFooter.svelte
```diff
@@ -0,0 +1,17 @@
...
+        </p>
+        <p class="underline">lino@up-csi.org</p>
+        <p>+63999999999</p>
+    </div>
+    <div class="flex flex-col text-right underline">
+        <a href="https://www.facebook.com/upcsi">Facebook</a>
+        <a href="https://twitter.com/officialupcsi">Twitter</a>
+        <a href="https://www.linkedin.com/company/up-center-for-student-innovations/">LinkedIn</a>
+        <a href="https://github.com/up-csi">GitHub</a>
+
```
BastiDood, line 14: Just a small nitpick: we can remove the extra space here. Don't worry, I can remove this on your behalf.
```suggestion
```

BastiDood (approved):
LGTM! 🚀

PR #9, opened 2023-12-13
feat: add navbar component for desktop layout

syncopascual:
This PR adds a navbar component for the website (designed for desktop layout). Note that in the Figma file, it looks like the text color of the other tabs are a shade gray. I couldn't pinpoint what shade was used in the design, so I kept them as the default inherited color for now.

### DIFF
--- a/src/lib/components/NavbarComponent.svelte
+++ b/src/lib/components/NavbarComponent.svelte
@@ -1,0 +1,28 @@
+<nav>
+    <ul class="flex flex-row gap-4 items-center bg-warm-white justify-between pl-8 pr-1 py-1 rounded-xl">
+        <li>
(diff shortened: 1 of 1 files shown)


BastiDood (changes requested):

src/lib/components/NavbarComponent.svelte
```diff
@@ -0,0 +1,37 @@
+<script>
+    import upcsi from '$lib/icons/upcsi.svg';
+</script>
+<nav>
```
BastiDood, line 4: Just a nitpick on the spacing here. Though, I don't mind committing this on your behalf. 👌
```suggestion
</script>

<nav>
```

BastiDood (approved):
LGTM! Great work on the navigation bar. 🚀

PR #10, opened 2023-12-13
feat: add cartoon card component

TechyChair:
Created cartoon card component and added its placeholder img in .svg format


BastiDood (approved):
Nice work! Just had to optimize some icons along the way.

PR #13, opened 2023-12-13
feat: add contact component

mkpautin:
This PR adds a contact component. Uses a different image of Lino since I wasn't able to find the one used on Figma.

### DIFF
--- a/src/lib/components/Contact.svelte
+++ b/src/lib/components/Contact.svelte
@@ -1,0 +1,14 @@
+<script>
+    import BlackButton from "./BlackButton.svelte";
+    import lino from "$lib/lino-hero.svg";
+</script>
+
+<section class="flex flex-col flex-wrap items-center sm:flex-row my-20">
+    <div class="sm:w-64 w-48 m-4">
+        <img alt="Cool Lino" src={lino} />
+    </div>
+    <div class="sm:text-left text-center sm:w-max w-3/5 px-5">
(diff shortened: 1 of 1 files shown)


src/lib/components/Contact.svelte
```diff
@@ -0,0 +1,16 @@
...
+
+<section class="my-20 flex flex-col flex-wrap items-center sm:flex-row">
+    <div class="m-4 w-48 sm:w-64">
+        <img alt="Cool Lino" src={lino} />
+    </div>
+    <div class="w-3/5 px-5 text-center sm:w-max sm:text-left">
+        <h1 class="mb-6 font-dm text-3xl font-light text-csi-black sm:text-5xl">
+            Let's innovate together!
+        </h1>
+        <button><BlackButton /></button>
```
BastiDood, line 14: The `BlackButton` component actually already contains a `<button>` already. This outer `<button>` is redundant.
```suggestion
        <BlackButton />
```

src/lib/components/Contact.svelte
```diff
@@ -0,0 +1,16 @@
+<script>
+    import BlackButton from './BlackButton.svelte';
```
BastiDood, line 2: For the sake of future-proofing the component (from being moved around), let's use absolute import paths here relative to `$lib` instead.
```suggestion
    import BlackButton from '$lib/components/BlackButton.svelte';
```

BastiDood (changes requested):

BastiDood (approved):
Thanks for this. LGTM! 🚀

PR #15, opened 2023-12-19
feat: add "Why UP CSI?" section to homepage & replace placeholder code in content container 

KindaExists:
This PR adds a "Why UP CSI?" section that contains two content container components (without cards) to the homepage. Additionally, this PR replaces the placeholder code in the content container component with a Svelte slot.


src/lib/components/ContentContainer.svelte
```diff
@@ -1,47 +1,25 @@
...
-    <div class="flex h-full flex-1 flex-col items-center gap-6 lg:flex-row lg:items-start">
-        <div class="flex w-full flex-1 flex-col items-center gap-4">
-            <img src={src1} alt={alt1} class="m-0 rounded-3xl bg-black text-csi-white" />
-            <p class="mx-2 my-0 text-center text-black">{imageText1}</p>
-        </div>
-        <div class="flex w-full flex-1 flex-col items-center gap-4">
-            <img src={src2} alt={alt2} class="m-0 rounded-3xl bg-black text-csi-white" />
-            <p class="mx-2 my-0 text-center text-black">{imageText2}</p>
+    <!-- Container for basic card components -->
+    {#if $$slots.cards}
```
BastiDood, line 20: This feels strange to me. I don't think we should be defensively checking against the existence of a slot prop. If the caller forgets to provide a child item, the `<slot name="cards" />` gets rendered as an empty element anyway.

src/routes/+page.svelte
```diff
@@ -53,6 +57,19 @@
         <li>Providing solutions to <em>your</em> problems.</li>
     </ul>
 </section>
+<section class="prose max-w-full">
+    <h1 class="font-dm">Why UP CSI?</h1>
+    <ContentContainer
+        topText="For organizations"
+        headerText="We offer services catered to your specific software needs."
+        bottomText="Automate the busy work in your day-to-day operations so people can get home on time more often."
```
BastiDood, line 65: Let's use [named slots](https://learn.svelte.dev/tutorial/named-slots) here instead. Normally, when we deal with content, it's better to treat them as slotted elements rather than props.

BastiDood (changes requested):

src/lib/components/ContentContainer.svelte
KindaExists: Hello! I added the check since the flexbox styling from `<div class="flex h-full flex-1 flex-col items-center gap-6 lg:flex-row lg:items-start">` would be applied regardless if the "cards" slot was empty or not. This means the right half of the content container will be empty space even if the slot is unused. In retrospect, it might be better to just remove the parent `<div>` entirely, keeping the `<slot name="cards" />`, and letting the slotted elements handle it instead.

BastiDood: Yup, I agree! When things get too specific at the slot-level, I usually take that as a sign that perhaps it should be the caller's responsibility to pass the element into the slot rather than us fiddling with the styles from _inside_ the component. I like it when components make as few assumptions as possible about its enclosing environment.

src/lib/components/ContentContainer.svelte
```diff
@@ -1,47 +1,9 @@
...
-
-    // eslint-disable-next-line init-declarations
-    export let alt1: HTMLImgAttributes['alt'];
-    // eslint-disable-next-line init-declarations
-    export let alt2: HTMLImgAttributes['alt'];
-</script>
-
 <div
-    class="flex flex-col items-center justify-between gap-12 rounded-3xl bg-warm-white px-8 py-8 md:px-32 md:py-16 lg:flex-row"
+    class="my-6 flex flex-col items-center justify-between gap-12 rounded-3xl bg-warm-white px-8 py-8 md:px-32 md:py-16 lg:flex-row"
```
BastiDood, line 2: We typically avoid introducing margins at the component-level. Instead, we delegate that responsibility to the caller (i.e., the parent component) for the greatest flexibility.
```suggestion
    class="flex flex-col items-center justify-between gap-12 rounded-3xl bg-warm-white px-8 py-8 md:px-32 md:py-16 lg:flex-row"
```

src/routes/+page.svelte
```diff
@@ -1,3 +1,7 @@
+<script lang="ts">
```
BastiDood, line 1: TypeScript isn't actually necessary here since we don't use any of its fancy syntax.
```suggestion
<script>
```

BastiDood (changes requested):
Sorry for the late review. Only had the time to check this out now. Anyway, we're almost there! Just a few nitpicks about the implementation.

BastiDood:
Nice! Just gotta resolve that one merge conflict. 🚀

BastiDood (approved):
Nice work! Thank you for your contributions.

PR #16, opened 2024-01-08
feat: changed email from <p> to <a> to redirect

rue-22:
This PR changes the email from *<p>* element to *<a>* element to redirect user when they click it.

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -14,6 +14,4 @@
 <main class="m-4 space-y-16 sm:m-8 xl:m-32 xl:my-16">
     <slot />
 </main>
-<footer class="p-8 sm:p-12 md:p-16">
-    <SidesFooter />
-</footer>
+<SidesFooter />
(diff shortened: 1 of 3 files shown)


BastiDood (approved):
Nice catch! Thanks for this. I totally missed this the first time. :rocket:

src/routes/+layout.svelte
```diff
@@ -13,26 +13,4 @@
...
-                href="https://www.linkedin.com/company/up-center-for-student-innovations/"
-                >LinkedIn</a
-            >
-        </li>
-        <li>
-            <a target="_blank" rel="noopener" href="https://github.com/up-csi ">GitHub</a>
-        </li>
-    </ul>
-</footer>
+<footer class="p-8 sm:p-12 md:p-16"></footer>
```
BastiDood, line 16: Let's put the `SidesFooter` here instead.

BastiDood (changes requested):

src/routes/+layout.svelte
```diff
@@ -14,25 +15,5 @@
...
-                rel="noopener"
-                href="https://www.linkedin.com/company/up-center-for-student-innovations/"
-                >LinkedIn</a
-            >
-        </li>
-        <li>
-            <a target="_blank" rel="noopener" href="https://github.com/up-csi ">GitHub</a>
-        </li>
-    </ul>
+    <SidesFooter />
```
BastiDood, line 18: We could actually just put the `<footer>` inside the `SidesFooter` component itself. Instead of having a `<div>` as the root, we can just use `<footer>` instead.

BastiDood (changes requested):

PR #17, opened 2024-01-08
feat: add basic cards to "Why UP CSI?" section in home page

TechyChair:
This PR adds the four basic cards to their respective content containers ("For Students" and "For Organizations") in the "Why UP CSI?" section of the home page. This PR also modified certain CSS classes in the Cartoon Card component and added slots for its header and text.


BastiDood (approved):
Thanks for this. LGTM! 🚀

PR #19, opened 2024-01-08
feat: add Navbar to layout.svelte

syncopascual:
This PR adds the Navbar component to the top of the page (tentatively without removing the previous Navbar element) and adds a margin to the new Navbar component :>


src/lib/components/NavbarComponent.svelte
```diff
@@ -4,7 +4,7 @@
 
 <nav>
     <ul
-        class="flex flex-row items-center justify-between gap-4 rounded-xl bg-warm-white py-1 pl-8 pr-1"
+        class="m-2 flex flex-row items-center justify-between gap-4 rounded-xl bg-warm-white py-1 pl-8 pr-1"
```
BastiDood, line 7: Let's avoid placing the `margin` at the component-level. Ideally, this is something we put at the page-level or the layout-level to give maximum flexibility to whoever is gonna use our component. For instance, it would be unfortunate if importing the component forced us to include the `m-2` margin.

src/routes/+layout.svelte
```diff
@@ -3,8 +3,12 @@
     import NavList from './NavList.svelte';
     import SidesFooter from '$lib/components/SidesFooter.svelte';
     import upcsi from '$lib/icons/upcsi.svg';
+    import Navbar from '$lib/components/NavbarComponent.svelte';
```
BastiDood, line 6: This import should be sorted above `NavList`. Also do consider renaming it to `NavBar` for consistency. Maybe we can even rename the file as well.
```suggestion
    import NavBar from '$lib/components/NavBar.svelte';
```

BastiDood (changes requested):
Don't forget to check the formatters and linters! 👍

BastiDood (approved):
Awesome! LGTM 🚀

PR #38, opened 2024-03-13
feat: add code for mission and vision statements

rue-22:
This PR adds an `about` folder inside the `routes` folder. Also, the code for Mission and Vision statements for the website is included inside the `about` folder since it is only used in the About page. Placeholder images were used in the meantime.


BastiDood (approved):
Great work! Thanks for this. 🚀

PR #39, opened 2024-03-13
feat: combine footers and add responsive footer styling

KindaExists:
This PR combines the `SidesFooter`, `CenterFooter`, and `Footers` components into a single `Footer` component with responsive styling that leverages flexboxes and media queries. Additionally, dark mode text coloring was added for the Footer component. 

Note that the "Site built with SvelteKit" text does not currently use the gray text color (`#A6A6A6`) that can be seen in the original design. This is because I'm not sure if I should be editing the `tailwind.config.js` file to add arbitrary colors (there may also be a preferred name for the color if it were added).


BastiDood (approved):
LGTM! 🚀

BastiDood:
Some future work that should be addressed in the next PR is the fact that dark mode currently results in `text-csi-white` on top of `bg-csi-white`, which makes it seem like the footer is invisible. Let's tackle the main `bg-` in the next PR. 👌

BastiDood:
Also, regarding the colors, feel free to add/edit the `tailwind.config.js` since I'll be reviewing them anyway. As long as the colors are consistent with the design system, it should be alright.

PR #40, opened 2024-03-14
feat: add Contact component to home page and adjust component styles

KindaExists:
This PR adds the `Contact` component to the home page and adjusts its styling to make it more compatible with mobile devices and smaller screens. 

Note that the container of the contact component has no background color yet as it is unclear to me if it is supposed to actually have one in the design. This is given that the container uses the same color as the main page background in dark mode and its page margins in the light mode version of the mobile design don't seem to follow the page margins used by other components in the rest of the design, as shown in the image below. 

image

Additionally, `lino-hero.svg` was used as a placeholder in place of the image of Lino wearing shades.


BastiDood (approved):
> Note that the container of the contact component has no background color yet as it is unclear to me if it is supposed to actually have one in the design.

I think generally speaking, we would prefer transparent backgrounds (despite what the Figma file shows). Aside from that, LGTM! 🚀

PR #41, opened 2024-03-15
feat: add gray color and dark mode to page background

KindaExists:
This PR adds the `gray` color (`#A6A6A6`) to the color themes and applies it to the "Built with SvelteKit" text in `Footer` (as well as the temporary attribution text). Also, this PR adds dark mode to the background of the page.


BastiDood (approved):
Oh, that was surprisingly straightforward... 😅

PR #42, opened 2024-03-17
fix(HeroComponent.svelte): tweak layout for slightly more consistency with Figma design

syncopascual:
This PR makes the sizing of the images and text layout a little more consistent with the Figma design. It also adds some dark mode styles (though doesn't use prose or prose-invert).

As a note for the next changes to be made for this component, the background color needs to be a gradient, and I think the image has to use some sort of (maybe?) absolute positioning to get the effect of Lino perching on the bottom of the Hero component.


BastiDood (approved):
Nice, thanks for this. LGTM! :rocket:

PR #44, opened 2024-03-22
feat: add carousel functionality

galierie:
This PR adds animation and autoplay functionality to the carousel component. However, there might be a bit of editing needed once there's official content for the "We've worked with" section, specifically for cases with 8 or less carousel cards.

### DIFF
--- a/src/lib/components/Carousel.svelte
+++ b/src/lib/components/Carousel.svelte
@@ -41,17 +41,17 @@
 </script>
 
 <div class="flex w-full justify-center gap-5 overflow-clip">
-    {#each shownPartners as partner (partner.id)}
+    {#each shownPartners as { id, name } (id)}
         <div animate:flip={{ duration: 500 }}>
-            <CarouselCard name={partner.name} />
+            <CarouselCard {name} />
         </div>
     {/each}
 </div>
 
 <div class="my-4 flex w-full justify-center gap-5 overflow-clip lg:hidden">
-    {#each shownPartnersMobile as partner (partner.id)}
+    {#each shownPartnersMobile as { id, name } (id)}
         <div animate:flip={{ duration: 500 }}>
-            <CarouselCard name={partner.name} />
+            <CarouselCard {name} />
         </div>
     {/each}
 </div>
--- a/src/lib/components/Carousel.svelte
+++ b/src/lib/components/Carousel.svelte
@@ -17,7 +17,7 @@
     ];
 </script>
 
-<div class="flex w-full justify-start gap-5 snap-x overflow-x-scroll">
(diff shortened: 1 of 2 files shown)


src/lib/components/CarouselCard.svelte
```diff
@@ -1,6 +1,11 @@
+<script>
+    /** @type {String} */
+    export let name = '';
+</script>
+
 <!-- Carousel Card with filler content -->
 <div class="w-56 flex-initial rounded-full bg-blue p-5 text-center font-inter">
     <!-- Sample Text -->
-    <p class="m-0">Lorem ipsum dolor sit amet consectetur</p>
+    <p class="m-0">{name}</p>
```
BastiDood, line 9: Let's use slots here to give maximum flexibility to the caller. Using slots will let them use bolded text, italics, links, etc.

src/lib/components/Carousel.svelte
```diff
@@ -1,11 +1,57 @@
...
+        </div>
+    {/each}
+</div>
+
+<div class="my-4 flex w-full justify-center gap-5 overflow-clip lg:hidden">
+    {#each shownPartnersMobile as partner (partner.id)}
+        <div animate:flip={{ duration: 500 }}>
+            <CarouselCard name={partner.name} />
+        </div>
+    {/each}
```
BastiDood, line 56: This is a little bit more concise.
```suggestion
    {#each shownPartnersMobile as { id, name } (id)}
        <div animate:flip={{ duration: 500 }}>
            <CarouselCard {name} />
        </div>
    {/each}
```

src/lib/components/Carousel.svelte
```diff
@@ -1,11 +1,57 @@
...
+            ...partners.slice(next, next + 1),
+        ];
+        next = (next + 1) % partners.length;
+
+        shownPartnersMobile = [
+            ...shownPartnersMobile.slice(1, shownPartnersMobile.length),
+            ...partners.slice(nextMobile, nextMobile + 1),
+        ];
+        nextMobile = (nextMobile + 1) % partners.length;
+    }, 2000);
```
BastiDood, line 39: This is nice and all, but for the v1, I think we should stick with a simplified [scroll snap](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_scroll_snap) for now.

BastiDood (changes requested):

src/lib/components/Carousel.svelte
galierie: Ay hello! Do I still implement carousel autoplay? Also, do I style the scrollbar?

BastiDood: I think we can drop the autoplay for now. And let's use the default scrollbar for this PR in the meantime.

PR #45, opened 2024-03-22
feat: adjust content container styling

galierie:
This PR adjusts the styling of the content container component to better fit the website figma. Note that the background colors for the containers are not yet working but are already formatted according to Tailwind's one-off arbitrary value specs; as such, the background colors for now are csi-white and csi-black for light and dark mode, respectively.


src/lib/components/ContentContainer.svelte
```diff
@@ -1,8 +1,12 @@
 <div
-    class="flex flex-col items-center justify-between gap-12 rounded-3xl bg-warm-white px-8 py-8 md:px-32 md:py-16 lg:flex-row"
+    class="flex flex-col items-center justify-between gap-12 rounded-3xl bg-[#f6f6f6] px-8 py-8 md:px-32 md:py-16 xl:flex-row dark:bg-[#1d1d1d]"
```
BastiDood, line 2: Let's edit the `tailwind.config.js` so that these hard-coded numbers are in the configuration itself.

BastiDood (changes requested):

tailwind.config.js
BastiDood, line 1: It would be nice to have [color stops](https://tailwindcss.com/docs/customizing-colors#color-object-syntax) here instead of explicit color names. Feel free to check out [UIColors](https://uicolors.app/create) to help automate the process. Let's aim for hex colors that are as similar as possible to the original design.

BastiDood (changes requested):
One more change to go. Note that this may involve editing more than one file in the entire codebase.

galierie:
uhhhh nyarks

BastiDood (approved):
Awesome! Great work on the color stops. LGTM! 🚀

PR #46, opened 2024-03-22
feat: create SectionHeaderTextContainer component

syncopascual:
This PR creates the SectionHeaderTextContainer component for the About page! Note that the buttons still don't include the functionality to scroll to the designated sections on the page and that the images are still also placeholders for the background color/image.


BastiDood (commented):
It would be nice to put this component on the page already. 👍

BastiDood (changes requested):
Oops, I forgot to click the "request changes" button. 😅

syncopascual:
Oh oops I think something's up with the way I set up the about page, hold on

src/routes/about/+page.svelte
```diff
@@ -0,0 +1,15 @@
...
+
+<section>
+    <SectionHeaderTextContainer />
+</section>
+<section>
+    <MissionVision />
+</section>
+<section>
+    <Contact />
+</section>
```
BastiDood, line 15: I like the use of `section` here, but I think they belong in the component level, not in the page level. Let's put the `section` as the root element (of the respective components) rather than the `div`.

BastiDood (changes requested):

src/routes/NavBar.svelte
```diff
@@ -16,7 +16,7 @@
             <img src={upcsi} alt="UP CSI Logo" class="h-8 object-contain" />
         </li>
         <li class="hidden lg:block"><a href="/">Home</a></li>
-        <li class="hidden lg:block"><a href="/#about">About</a></li>
+        <li class="hidden lg:block"><a href="/about">About</a></li>
```
BastiDood, line 19: While we're at it, let's also remove the corresponding `id="about"` bookmark.

BastiDood (approved):
Awesome work here. LGTM! :rocket:

PR #47, opened 2024-03-23
feat: adjust navbar styling

galierie:
This PR adjusts the styling of the navigation bar component to better fit the website figma. Note that this PR also edits tailwind.config.js to add dark mode toggle functionality.


BastiDood (approved):
LGTM! Future work will involve componentizing the mobile menu so that the `NavBar` component doesn't become too cluttered. For now, though, we can merge this. Thanks! :rocket:

> [!NOTE]
> Please take note of the refactors I did to the theme store. It's not a good idea to create globally initialized stores in a SvelteKit server because this global instance leaks across all HTTP requests. Ideally, we should construct a brand new store per request to make sure no state gets leaked. See the [documentation](https://kit.svelte.dev/docs/state-management#using-stores-with-context) for more details.

PR #49, opened 2024-04-01
feat: add basic prose-invert fix various colors

KindaExists:
This PR adds `prose-invert` and configures typography themes to provide (very) basic dark mode coloring to various pieces of text that had the `prose` class. Note that I only specified typography prose colors for body text, heading text, and bullet points as this seems to be all that is needed at the moment. In the future, other contributors will likely have to add or change these color themes depending on what the original design needs.

Additionally, the color gradient was implemented into `HeroComponent.svelte`. For this, I had to change the `tailwind.config.ts` such that `backgroundImage` theme extends instead, as replacing tailwind's default `backgroundImage` theme means removing its ability to generate gradients (tailwind background gradients are defined under `backgroundImage`). 

Besides those, this PR just tweaks the colors and styles of a few components for added consistency with the design and uniformity with other components.


BastiDood (approved):
Nice! I don't mind the configuration change. Great work on that. LGTM! 🚀

PR #50, opened 2024-04-03
feat: remove duplicate content from Home page

syncopascual:
This PR removes duplicate content from the Home page and fixes minor text and element alignment issues


BastiDood (approved):
Awesome! Glad to see the work in the other components being integrated here. 🚀

PR #51, opened 2024-04-07
fix: fix lino and navbar positioning

KindaExists:
This PR adjusts various styles in `HeroComponent.svelte` and fixes the position of Lino in the Hero component so that it looks more like Lino is perched on the border of the component for added consistency with the design. Also, the navbar was changed to have `position: fixed` when viewed on large screens so that the navbar does not affect normal document flow and matches the design (where the navbar starts above the hero component).

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -10,10 +10,10 @@
     $: if (browser) document.documentElement.classList.toggle('dark', $theme);
 </script>
 
-<header class="sticky top-0 lg:top-12">
+<header class="sticky lg:fixed top-0 lg:top-10 lg:inset-x-0 z-50">
     <NavBar />
 </header>
-<main class="m-4 space-y-16 sm:m-8 xl:m-32 xl:my-16">
+<main class="m-4 space-y-16 sm:mt-5 sm:m-8 xl:mx-32 xl:mb-16">
     <slot />
 </main>
 <Footer />
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -10,10 +10,10 @@
     $: if (browser) document.documentElement.classList.toggle('dark', $theme);
 </script>
 
-<header class="sticky lg:fixed top-0 lg:top-10 lg:inset-x-0 z-50">
+<header class="sticky top-0 z-50 lg:fixed lg:inset-x-0 lg:top-10">
     <NavBar />
 </header>
-<main class="m-4 space-y-16 sm:mt-5 sm:m-8 xl:mx-32 xl:mb-16">
+<main class="m-4 space-y-16 sm:m-8 sm:mt-5 xl:mx-32 xl:mb-16">
     <slot />
 </main>
 <Footer />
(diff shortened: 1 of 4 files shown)


tailwind.config.ts
```diff
@@ -53,6 +53,13 @@ export default {
             backgroundImage: {
                 lino: 'url("$lib/lino-hero.svg")',
             },
+            height: {
+                '136': '34rem',
+            },
+            width: {
+                '104': '26rem',
+                '156': '39rem',
+            },
```
BastiDood, line 62: These seem like one-off arbitrary values rather than something we should encode into the Tailwind configuration. I think the inline `[...]` style syntax is more appropriate here.

src/routes/+layout.svelte
```diff
@@ -10,10 +10,10 @@
     $: if (browser) document.documentElement.classList.toggle('dark', $theme);
 </script>
 
-<header class="sticky top-0 lg:top-12">
+<header class="sticky top-0 z-50 lg:fixed lg:inset-x-0 lg:top-10">
```
BastiDood, line 13: I'm still on the fence about this because `sticky` already decays to the `fixed` positioning once we start scrolling. With that said, I personally don't mind the navbar affecting the document flow because navbars are typically space-occupying components than floating widgets (e.g., a chat bot conversation UI).

BastiDood (changes requested):

src/routes/+layout.svelte
KindaExists: Honestly, not sure about this change either. As said on the PR, I changed it to `fixed` positioning mainly because the large screen design had the navbar overlapping with the Hero Component and I was trying to make them look more similar to each other. So, I suppose it really depends on whether we want to follow this aspect of the original design or deviate from it. Personally, I think either way is fine, given this change is relatively inconsequential to begin with.

BastiDood: I think we can deviate from it in this instance. The overlap feels a little too cramped at the top. 👍

Also, while we're fixing the navbar, could you also add a [`shadow`](https://tailwindcss.com/docs/drop-shadow) around it just so the component is more readable? Feel free to choose whichever `sm`, `md`, `lg`, etc. variant that looks best but most subtle at the same time.

KindaExists:
Reverted the fixed positioning, and added a minor box-shadow to the navbar (used `shadow` over `drop-shadow` since it appears to be more flexible). I ran into an issue while adding the `shadow` since it doesn't do much in terms of making the navbar more distinguishable in dark mode, I tried playing around with several shadow colors but I personally didn't like how it looked so I didn't push that change for the moment. I think it would be a good idea to increase the contrast between the navbar and the background in dark mode by giving the navbar a lighter/darker background color, this should also indirectly help the `shadow` seem more apparent and further distinguish the navbar from the background.

BastiDood (approved):
> If we want it to be more distinguishable in dark mode, I think it would be a good idea to increase the contrast between the navbar's background-color and that of the background by giving the navbar a different background color. This should also indirectly help the shadow seem more apparent and further distinguish the navbar from the background.

I'm totally open to receiving a PR on that if you want to play with the colors. Although, I must say that it does make sense that the shadows are not too seen in dark mode because there can be no shadows in the absence of light. 😅 That's a non-issue in my book, but I would nevertheless appreciate efforts to choose a more "contrast-y" color palette.

With that aside, this already looks great. Let's defer that work to another PR. LGTM! 🚀

PR #53, opened 2024-04-21
feat: remove Carousel from home page

galierie:
This PR removes the "We've worked with" section on the CSI home page.


BastiDood (approved):
Thanks for this! Must've been really hard to implement. 😅

PR #54, opened 2024-04-21
fix: remove Navbar bugs

galierie:
This PR fixes faulty links on the Navbar. This PR also fixes the behavior of the Mobile Navbar as it has come to my attention that people think that the links on the Mobile Navbar are faulty as well.


BastiDood (approved):
Nice catch on this. That was my refactor that caused the bug. LGTM! 🚀

PR #56, opened 2024-05-04
feat: add event cards panel

KindaExists:
This PR adds an `EventPanel` that contains `EventCard` components and has pagination and card filtering functionalities. The information and details used for the cards were placed in `events.json` to allow for looping and ease of editing. Images used for the cards were kept in the `/static/thumbnails/events/` folder for the sake of looping as well. Placeholder/filler text was used to fill the card details for the moment and images used were taken from the [B&C drive](https://drive.google.com/drive/folders/1RFpQRa-rLEOsUbIkLwkWOr6mUP-JPavM).

Note that, in this implementation of the [design from the mockup](https://www.figma.com/file/gdObjB04j7BT36Q0lQ9GzL/UP-CSI-Website?type=design&node-id=203%3A16&mode=design&t=JuSvrpPccS4B64M0-1), I attempted to match the styles of the new components with the rest of the components already implemented (i.e. rounded corners, shadows, etc.). Also, note that the pagination functionality was implemented entirely through the use of state for simplicity, although I'm not sure if this is the proper way of implementing it. Moreover, the pagination component does not currently account for the possibility of there being too many pages that it overflows, but I believe it would be better to address this in a future PR if this one is accepted.


BastiDood (approved):
I'm super impressed by this PR! Awesome work on this. We can merge this now, but I'd like to raise some notes on future work:

1. I'd imagine that this pagination component you wrote would be immensely valuable to other pages as well (i.e., the projects and portfolio). @syncopascual Let's have someone assigned to refactor the pagination logic into its own component eventually.
2. The event cards could use some more padding. That's just a nitpick, though. Let me know which padding _feels_ better for you.
3. As noted by Steve Schoger in his 2019 talk on ["Refactoring UI"](https://www.youtube.com/watch?v=7Z9rrryIOC4), de-emphasizing subtitles and sub-headings goes a long way in establishing a visual hierarchy in text. Future PRs should consider using neutral grays on the "less important" text.

PR #58, opened 2024-05-12
feat(about): add team component

KindaExists:
This PR adds the "Team Component" (`TeamPanel.svelte`) to the `about` page with committee filtering functionality. This component was implemented in a similar manner to the `EventPanel.svelte` from the previous PR aside from the lack of pagination (this wasn't included in the Figma design). Repeating what I said in the last PR, the information of each team member that was used for generating `TeamCard`s were placed in `team.json` for ease of editing, and images used for the cards were kept in the `/static/about` directory for the sake of looping. 

Admittedly unconventional usage of CSS grids was done to achieve the overlay effect shown on cards on the Figma design instead of using `background-image` (`bg-[...]`). This was mainly done to avoid removing the `` tag and its `alt` text which may be useful if we decide to work on accessibility later. 

Moving onto some notable issues in the implementation of this PR.

Firstly, there are currently no black social media icons for light mode (for small screen devices). While this is admittedly minor and would likely not be difficult to solve, it may be a good idea to either add an icon library or create an icon loader svelte script to store and handle the svgs instead of handling them manually (or turning them into `.svelte` files). I think these could make icons less tedious to handle for every use case in the future especially if we end up adding a lot of them. That said, I can also add some temporary dark mode social media icons (i.e. `instagram-dark.svg`, `github-dark.svg`, etc.) if needed before pushing this PR (assuming it gets accepted in the first place). Additionally, I only added 3 social media icons for the moment (github, linkedin, instagram), with `link.svg` as a fallback icon.

Next, the way I implemented the overlay for the cards may have several issues at the moment. For one, since it uses `on:mouseeneter` and `on:mouseleave`, I have yet to check if it would work for mobile/touchscreen devices so this may need to be adjusted later on. Second, due to the use of an `if-else` logic block, the card overlay will reload every time it is hovered over which might not be the preferred behavior. A possible alternative solution is to instead hide the overlay through CSS, but I did not choose to use this method for the moment for the sake of simplicity and since I'd like to first hear thoughts on which would be preferable for this case. Additionally, it should be noted that using the hiding method in CSS will, to my knowledge, remove the possibility of adding transitions in the future (unless we use `invisible`). 

Lastly, the implementation of the cards means that in small/mobile screens, the events will still be firing and editing the DOM even if they have no visible effect due to being hidden. I'm not exactly sure what this would imply in terms of performance but this is still something to note. 

_Sidenote: Surprisingly, this component ended up being more tedious and messy to implement than the last, even with all the things made in the previous PR. Also, don't mind the fact that this PR was written at this ungodly hour._


BastiDood (approved):
Hello there! I cleaned up the implementation a bit. Instead of using dynamic JSON for configuration, I just inlined the imports so that Vite can still detect them in the dependency tree.

Moreover, I've also added a new icon library that we can use for all icons moving forward. See [`@steeze-ui/icons`](https://github.com/steeze-ui/icons) for more information.

PR #61, opened 2024-07-04
feat: add featured event card in events route

rue-22:
I added the featured event card that is found in the events route of the website. I've used slots to pass the image, event name, and the description to the component. On the other hand, I've used props (array of strings) to pass the tags and schedules for the card.


BastiDood (changes requested):
Cool, can we add this component to the page in this PR? On another note, please do update your branch to the latest. Thanks!

PR #62, opened 2024-07-07
feat: replace placeholder cartoons in Home and About page

syncopascual:
This PR replaces images in the Home and About page with more fitting cartoons. The only remaining placeholder is the Lino at the top of both pages, which in the Figma file is an image of the Oblation (I can't seem to find this exact picture, so I left it as Lino for now).


src/lib/components/CartoonCard.svelte
```diff
@@ -1,13 +1,17 @@
 <script>
-    import src from '$lib/cartoon-card-placeholder.svg';
+    export let src = '$lib/lino-sablay.svg';
```
BastiDood, line 2: I think it's better if we explicitly require the caller to provide the `src` rather than defaulting and assuming that `lino-sablay.svg` is what the caller meant to put.
```suggestion
    // eslint-disable-next-line init-declarations
    export let src: string;
```

src/routes/about/Card.svelte
```diff
@@ -1,10 +1,13 @@
 <script>
-    import src from '$lib/lino-sablay.svg';
+    export let src = '$lib/lino-sablay.svg';
+    export let bgStyles = 'bg-csi-neutral-200 rounded-t-full rounded-b-xl';
```
BastiDood, line 3: I believe this isn't necessary at the moment. Let's reintroduce it only when needed. 👌 
```suggestion
```

src/routes/about/Card.svelte
```diff
@@ -1,10 +1,13 @@
 <script>
-    import src from '$lib/lino-sablay.svg';
+    export let src = '$lib/lino-sablay.svg';
```
BastiDood, line 2: Same thoughts as the previous suggestion regarding defaulted `src` props.

src/lib/components/CartoonCard.svelte
```diff
@@ -1,13 +1,17 @@
...
+    <div
+        class="flex aspect-square items-center overflow-y-hidden rounded-3xl bg-csi-blue px-2 pt-5"
+    >
         <img {src} alt="Corporate Cartoon" />
     </div>
-    <div class="leading-6">
-        <h1 class="text-center font-dm text-3xl"><slot name="title" /></h1>
-        <p class="text-center text-csi-black dark:text-csi-white"><slot /></p>
-    </div>
+    {#if $$slots.title || $$slots.default}
```
BastiDood, line 11: Just curious, why did we go for `||` here instead of `&&`? In the body of conditional, we use both the `title` and `default` slot, so I was wondering why `||` is sufficient in this case.

BastiDood (changes requested):

src/lib/components/CartoonCard.svelte
syncopascual: Ah, I think the main reason was that in the Home +page.svelte where the `CartoonCard` components are called, I couldn't determine if the `title` slot was being used at all, so I just defaulted to `||` instead of checking for both slots being accessed. Would it be better to remove the element that uses the `title` slot entirely instead?

BastiDood: Yeah, if it's unused anyway, I think that's a fair direction we can take. I'm fine with removing it. 👍

BastiDood (changes requested):

BastiDood (approved):
Thanks for this. LGTM! 🚀

PR #63, opened 2024-07-08
feat: create Projects page skeleton

syncopascual:
This PR abstracts the EventCard and EventsPanel components for general use on the Projects page as well. I haven't renamed the files for the actual components yet, since I wasn't sure what would be a good indicator that it's used for both projects and Events.


BastiDood (commented):
Sorry for the delay here. I had been busy in the past few days. I like the generalization of the components here.

Though perhaps one nitpick that I'd like to say is that the JSON for the events should probably not be embedded in the component itself. What are your thoughts on moving the hard-coded stuff into a separate JSON file and importing them into the component instead?
```jsonc
// events.json
[{ ... }, { ... }, ...]
```
```svelte
<script>
    // Component.svelte
    import events from './events.json';
</script>
```

ehrelevant:
> Though perhaps one nitpick that I'd like to say is that the JSON for the events should probably not be embedded in the component itself. What are your thoughts on moving the hard-coded stuff into a separate JSON file and importing them into the component instead?
> 
> ```js
> // events.json
> [{ ... }, { ... }, ...]
> ```
> 
> ```svelte
> <script>
>     // Component.svelte
>     import events from './events.json';
> </script>
> ```

Heya! I kinda just read this PR right now and thought I should mention that my initial implementation for the events panel (#56) used a separate `events.json` file. If I'm not mistaken, I ran into an issue with this implementation since it would require images to be stored in `static` instead of `lib` (or use Vite's dynamic imports). Hence, we ended up embedding it into the component (#58). With that said, I realize that instead of putting the event data into a separate `json` file, we could put it in a separate `js` file. This way, we should be able to import from `lib` while keeping a separation between the component and event data.

```js
// events.js
import ascenc10n from '$lib/events/ascension.webp';
import bootcamp10 from '$lib/events/bootcamp-10.webp';

export const events = [{ ... }, { ... }, ...]
```

```svelte
<script>
    // Component.svelte
    import events from './events.js';
</script>
```

BastiDood:
Yes, I love this idea! Great use of Vite's static imports. 🚀

BastiDood:
Hello there! Just following up on this. It would be awesome to proceed with @ehrelevant's idea here.

BastiDood (approved):
Awesome work! I applied some of my refactors on top where each JSON data type is now defined within the `EventsPanel.svelte` component. LGTM! 🚀

Some notes for future work, though:
* Maybe we should reconsider `EventsPanel` as the component name. Being used as `ProjectsPanel`, we should probably rename it to something more general.

VeeIsForVanana:
It is amusing how this was literally the first notification I saw waking up LMAO. Thanks for the help @BastiDood !!! 🚀🚀🚀

PR #60, opened 2024-06-21
feat: add custom 404 error page

galierie:
This PR adds a custom 404 error page layout. This PR also adds a general page layout for unexpected errors.

PR #64, opened 2024-07-17
fix: create `404.html` page on build for github pages

ehrelevant:
At the moment, the custom 404 error page made in #60 will not appear if you attempt to go to a non-existent page (i.e. [https://up-csi.org/not-a-real-url](https://up-csi.org/not-a-real-url)). This is because the 404 error page displayed by github pages depends on a `404.html` which wasn't being generated during building. This PR attempts to fix this issue by adding `fallback: '404.html'` in `svelte.config.js` as suggested in the [SvelteKit documentation for github pages](https://kit.svelte.dev/docs/adapter-static#github-pages).

EDIT: fixed link formatting

### DIFF
--- a/svelte.config.js
+++ b/svelte.config.js
@@ -5,5 +5,9 @@ import { vitePreprocess } from '@sveltejs/vite-plugin-svelte';
 export default {
     extensions: ['.svelte', '.md'],
     preprocess: vitePreprocess(),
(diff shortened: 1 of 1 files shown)


BastiDood (approved):
Oh wow, I can't believe I missed that. Thanks for this! 🚀

Issue #52, opened 2024-04-21
Fix NavBar href for Projects page, Remove unfinished features


syncopascual:
Assigned to @galierie!

This is just a list of miscellaneous fixes for things that are unfinished/have to be removed for logistical reasons:

- [x] Remove carousel component from the page for now (#53)
- [x] Remove newsletter page from the NavBar (#54)
- [x] Fix Projects page href for the NavBar (for both desktop and mobile versions) (#54)

BastiDood:
Completed in #53 and #54.

PR #67, opened 2024-09-04
fix: remove Carousel section remnants

ehrelevant:
This PR removes some leftover remnants of the Carousel section on the home page for the time being, as per #52.


BastiDood (approved):
Nice catch. Thanks! 🚀

PR #69, opened 2024-09-10
feat(Events): add featured event card to page

ehrelevant:
This PR was built on top of #61, adding the event card created by @rue-22 to the page alongside some placeholder content and basic dark mode compatibility.

**EDIT(@BastiDood):** Closes #61.


src/routes/events/FeaturedEvent.svelte
```diff
@@ -0,0 +1,47 @@
+<script lang="ts">
+    import Button from '$lib/components/Button.svelte';
+
+    export let tags: string[] = [];
+    export let schedule: string[] = [];
```
BastiDood, line 5: I typically discourage defaulted props in any framework because it's so easy to forget that they affect the component. As a best practice, let's force the caller to always pass in the array.

```suggestion
    // eslint-disable-next-line init-declarations
    export let tags: string[];
    // eslint-disable-next-line init-declarations
    export let schedule: string[];
```

On a separate note, though, perhaps we could also use a `const enum` somewhere so that the `tags` aren't just the generic `string[]`. In terms of type safety, I'm happy to see stricter literals here.

src/routes/events/FeaturedEvent.svelte
```diff
@@ -0,0 +1,47 @@
...
+        <slot name="image" />
+    </div>
+
+    <div class="flex w-2/3 flex-col items-center gap-6 md:items-start">
+        <!-- Tags -->
+        <div class="flex gap-4">
+            {#each tags as tag}
+                <Button>
+                    {tag}
+                </Button>
```
BastiDood, line 24: Just a nitpick; the extra lines are kinda overkill. 😅
```suggestion
                <Button>{tag}</Button>
```

src/routes/events/FeaturedEvent.svelte
```diff
@@ -0,0 +1,47 @@
...
+            {/each}
+        </div>
+
+        <!-- Event Name -->
+        <h2 class="font-inter text-4xl font-bold"><slot name="name" /></h2>
```
BastiDood, line 29: Are you sure about the self-closing tag here?
```suggestion
        <h2 class="font-inter text-4xl font-bold"><slot name="name"></h2>
```

src/routes/events/FeaturedEvent.svelte
```diff
@@ -0,0 +1,47 @@
+<script lang="ts">
+    import Button from '$lib/components/Button.svelte';
+
+    export let tags: string[] = [];
+    export let schedule: string[] = [];
+</script>
+
+<!-- Featured Card Event
+tags and schedule - pass as props
+image, event name, description - pass as slot -->
```
BastiDood, line 10: Looks like you were trying to do an [`@component`](https://svelte.dev/docs/faq#how-do-i-document-my-components) doc-comment here.

BastiDood (changes requested):

src/routes/events/FeaturedEvent.svelte
ehrelevant: Ah should clarify that I didn't modify the original component from #61 much besides adding dark mode styles. My main goal for this PR was primarily to add the existing component to the page. Still, will go over the requested changes as soon as I can.

src/routes/events/FeaturedEvent.svelte
ehrelevant: I've updated my branch to implement your suggestions. I also added some `const enum`s for Tags, State, and Type in `events.ts` and used them in some components. However, I couldn't really use them in many places, since I couldn't change the `Event` type in the `EventsPanel` due to both the Events and Projects pages depending on it. With that said, I believe it will be necessary for us to rethink the implementation of tag filters when generalizing `EventsPanel` to include the type safety that `enum`s provide to some degree (using unions possibly).

BastiDood (approved):
I love it! Excellent work. LGTM! 🚀

src/routes/events/FeaturedEvent.svelte
BastiDood: Sure! We can do this in a follow-up PR. In the meantime, your implementation works great. 👍

PR #72, opened 2024-09-15
"Added functionality to buttons on homepage and navigation bar"

LantingKurt:
- Button **_See About_** leads to **About** page
- Both buttons **_Get in Touch_** on the home page lead to the UP CSI’s contact information on the footer

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -104,7 +104,10 @@ export default tsEslint.config(
             'prefer-template': 'error',
             radix: 'error',
             'require-await': 'error',
-            'sort-imports': ['error', { allowSeparatedGroups: true }],
+            'sort-imports': [
+                'error',
+                { ignoreCase: true, ignoreDeclarationSort: false, allowSeparatedGroups: true },
+            ],
             'spaced-comment': ['error', 'always', { markers: ['/'] }],
             yoda: ['warn', 'never', { exceptRange: true }],
         },
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
         "@steeze-ui/simple-icons": "^1.9.0",
         "@steeze-ui/svelte-icon": "^1.5.0",
         "@sveltejs/adapter-static": "^3.0.4",
-        "@sveltejs/kit": "^2.5.26",
+        "@sveltejs/kit": "^2.5.27",
         "@sveltejs/vite-plugin-svelte": "^3.1.2",
         "@tailwindcss/forms": "^0.5.9",
         "@tailwindcss/typography": "^0.5.15",
@@ -36,7 +36,7 @@
         "cssnano": "^7.0.6",
         "eslint": "^9.10.0",
         "eslint-config-prettier": "^9.1.0",
-        "eslint-plugin-svelte": "^2.43.0",
+        "eslint-plugin-svelte": "^2.44.0",
         "globals": "^14.0.0",
         "htmlnano": "^2.1.1",
         "postcss-prune-var": "^1.1.2",
@@ -47,13 +47,13 @@
         "stylelint-config-standard": "^36.0.1",
         "stylelint-config-tailwindcss": "^0.0.7",
         "svelte": "^4.2.19",
-        "svelte-check": "^4.0.1",
-        "svelte-eslint-parser": "^0.41.0",
+        "svelte-check": "^4.0.2",
+        "svelte-eslint-parser": "^0.41.1",
         "tailwindcss": "^3.4.11",
         "terser": "^5.32.0",
         "typescript": "^5.6.2",
         "typescript-eslint": "^8.5.0",
-        "vite": "^5.4.4"
+        "vite": "^5.4.5"
     },
     "dependencies": {
         "smooth-scroll": "^16.1.3"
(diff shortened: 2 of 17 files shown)


src/lib/components/HeroComponent.svelte
```diff
@@ -1,6 +1,12 @@
 <script>
     import Button from './Button.svelte';
     import src from '$lib/lino-sablay.svg';
+    function scrollToContact() {
+        const contactSection = document.getElementById('contact-section');
```
BastiDood, line 5: In Svelte, we discourage selecting an element by its ID. Instead, let's prefer yung [`bind:this`](https://svelte.dev/docs/element-directives#bind-this). Then let's put the browser-only logic inside an [`onMount`](https://svelte.dev/docs/svelte#onmount) function so that it doesn't fail during the build process.

src/routes/WhoWeAre.svelte
```diff
@@ -16,7 +17,7 @@
         </p>
         <div class="w-full text-center md:text-left">
             <Button>
-                <span>See About</span>
+                <button on:click={() => goto('/about')}>See About</button>
```
BastiDood, line 20: For accessibility reasons, let's not use `button` to emulate what an `<a href>` is already meant to do by default.

BastiDood (changes requested):
I have some major concerns in the current implementation. Kindly resolve the comments below.

src/routes/WhoWeAre.svelte
BastiDood: Here is how I would refactor this:
```suggestion
                <a href="/about/"}>See About</button>
```

vite.config.js.timestamp-1726409402364-b33a72027e561.mjs
BastiDood, line 1: This was accidentally committed into source control. Kindly remove this file.

BastiDood (changes requested):
It just occurred to me now that what we're trying to achieve here is a smooth scroll. In that case, I propose the following changes:

```html
<!-- The ID is necessary here as the bookmark target. -->
<a href="/#about">See About</a>
<section id="about"></section>
```
```html
<!-- app.html -->
<html class="scroll-smooth">
</html>
```

This native CSS solution is the most elegant way to do this with zero JavaScript required (i.e., no Svelte stores necessary).

* [CSS `scroll-behavior` by MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior)
* [Scroll Behavior by Tailwind CSS](https://tailwindcss.com/docs/scroll-behavior)
* [How to Create a Smooth Scroll by w3schools](https://www.w3schools.com/howto/howto_css_smooth_scroll.asp)

BastiDood:
@LantingKurt Here is an additional resource on HTML bookmarks, which implements the exact behavior you tried to emulate in JavaScript: [HTML Links - Create Bookmark](https://www.w3schools.com/html/html_links_bookmarks.asp).

eslint.config.js
```diff
@@ -104,7 +104,10 @@ export default tsEslint.config(
             'prefer-template': 'error',
             radix: 'error',
             'require-await': 'error',
-            'sort-imports': ['error', { allowSeparatedGroups: true }],
+            'sort-imports': [
```
BastiDood, line 107: What's with the relaxation of the rules? These were purposely (overly) strict to enforce convention so that there is no need to discuss formatting in PRs.

package.json
```diff
@@ -47,12 +47,15 @@
...
+        "svelte-eslint-parser": "^0.41.1",
         "tailwindcss": "^3.4.11",
         "terser": "^5.32.0",
         "typescript": "^5.6.2",
         "typescript-eslint": "^8.5.0",
-        "vite": "^5.4.4"
+        "vite": "^5.4.5"
+    },
+    "dependencies": {
+        "smooth-scroll": "^16.1.3"
```
BastiDood, line 59: This dependency is no longer needed with the native CSS solution that I proposed.

VeeIsForVanana:
I also have a few issues with this PR, aside from @BastiDood 's issues as raised above.

In any case, congratulations @LantingKurt on your first work (as far as I know) on the CSI Website. We just need to do some cleaning up.

First of all - a point on style, the commits touch on too many things at once. You could have separated out the commits for implementing the "Get in Touch" and "See About" buttons. This would make it easier to debug any issues relating to either individually. 

Second style point: your commits do not follow our **conventional commit standard**. Please read [this gist](https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13). In the future, I especially expect adherence to the `types` and `description` standards. 

On a more implementation-specific point, is it not possible for 

```svelte
<section bind:this={$contactSection} class="prose max-w-full dark:prose-invert">
    <Contact />
</section>
```

To be added to the bottom of the root `+layout.svelte` instead of being repeated across multiple pages?

ehrelevant:
I'll be closing this PR since most of the features meant to be added in this PR have already been addressed by #83 and #89. Still, a link to the About page is currently missing but we can probably incorporate that while addressing #88.

PR #76, opened 2024-09-23
feat(Team): move Team component to people page

### LINKED ISSUE #74: Move TeamComponent to People Page

ehrelevant:
This PR closes #74. Going over the changes:
- Move `TeamPanel` and `TeamCard` from the "about" page to the "people" page since having it in "about" seemed strange.
- Put team details into a separate `team.ts` file, similar to what we did for `EventsPanel` (#63).
- Fix the issue with social media icons being difficult to see in light mode (to be in line with original figma design).
- Add fade transitions when switching in and out of Team card overlays
  - This was done for both stylistic and practical reasons, as it gives leeway from `mouseleave` events from not being triggered
- Add various other transition and hover effects

I should also bring this up since it differs slightly from my suggestion back in #63:

```ts
const team: Member[] = [
  // ...
];

export default team;
```

In retrospect, doing a direct default export (i.e. `export default [ /* ... */ ] satisfies Member[]; `) means that typing information won't be recognized by the editor when it is imported into another file. This makes things a bit cumbersome when working with the imported object since code completion and type hints are not going to appear. There are likely other ways to get around it, but I think this is more appropriate overall. This shouldn't have an effect on `events.ts` and `projects.ts`, but it may be something worth taking note of when changing the implementation of those files or `EventsPanel`. 

Besides that, there's not much else to go over for this PR, just a few basic changes.


src/routes/people/team.ts
VeeIsForVanana, line 1: Good separation of the data and presentation for the Teams. I think this makes it better for when we start having the full membership initiate PRs into the website.

VeeIsForVanana (commented):
LGTM. 

My only real comment is on how extensible the `team.ts` approach will be when we start asking the membership to add themselves to the website via PRs. I'm wondering if we could prepare a folder under `people` where they could just add JSON files, then have `team.ts` collate all of them at once and append them all into the `team` list. How does that sound?

ehrelevant:
> I'm wondering if we could prepare a folder under `people` where they could just add JSON files, then have `team.ts` collate all of them at once and append them all into the `team` list. How does that sound?

Ah, I was supposed to suggest something similar to that actually, but it slipped my mind when I wrote the PR. Though, for that idea, I was originally just thinking of having separate `ts`/JSON files per committee. That said, I think your idea would be much better overall, although maybe we should place this folder somewhere in `lib` rather than in the `people` folder so as to keep the `routes` relatively clean.

VeeIsForVanana (approved):
Got it. Let's raise it to an Issue for the moment, to be done later down the line. Otherwise, approved.

PR #78, opened 2024-09-27
fix: add People page link to mobile navbar

galierie:
This PR adds a missing page link (People) to the mobile navbar.


BastiDood (approved):
I believe I originally deleted this because the website was incomplete at the time, but still featured real faces. I didn't want to give the false impression that the best that these "faces" could present is an incomplete web page, so I removed the link. Now that the website is more complete, though, I think this is fine.

VeeIsForVanana (approved):
All clear on this. LGTM! 🚀🚀🚀

PR #85, opened 2024-10-03
feat: generalization of events and projects components

syncopascual:
Pushing @ehrelevant's changes to generalize the events/projects pagination and tag logic. Here's just a general summary of the changes made in that branch and some other error/bug fixing and chores I added to try cleaning some stuff up (though there wasn't too much to add on to!):

- Generalized the behavior of the panel that contains the individual events/projects into `FilterPagePanel` rather than the projects page just using the EventsPanel component
- Created a new `ProjectsCard` component to be used in the `FilterPagePanel` rather than projects reusing EventsCard components
- Moved events and projects typescript files to `lib`

As an aside, an error pops up in `FilterPagePanel.svelte` when trying to set `this={cardComponent}` while rendering the individual. I got rid of the eslint error for now by just checking for the `state` and `dates` prop to choose which kind of component to render, but I'm not sure if this is the best way to do this.


src/lib/components/FilterPagePanel.svelte
```diff
@@ -80,15 +72,12 @@
...
-                    <svelte:fragment>
-                        <p class="m-0">{state} - {type}</p>
-                        <h2 class="m-0">{title}</h2>
-                        <p class="m-0">{dates}</p>
-                        <p class="m-0 grow overflow-hidden">{description}</p>
-                    </svelte:fragment>
-                </EventCard>
+            {#each filteredCardsInPage as cardProps}
+                {#if 'state' in cardProps && 'dates' in cardProps}
+                    <svelte:component this={cardComponent} {...cardProps} />
```
ehrelevant, line 77:
```suggestion
                {#if 'state' in cardProps}
                    <svelte:component this={EventCard} {...cardProps} />
```
We don't need both `'state' in cardProps` and `'dates' in cardProps` to satisfy the type checker. Just one of these would do.

Also, I suppose rather than `state` and `dates`, it might be better to change `title`—in both the `Event` and `Project` interfaces—to `event` and `project`, then base it off that.

Besides that, I'm still stumped over other possible ways we could've implemented this without the type-checker complaining. I think that if we wanted to do this properly, then we would have to redo this component entirely. To be honest, I don't think this is worth dealing with at the moment.

src/lib/components/ProjectCard.svelte
```diff
@@ -0,0 +1,24 @@
...
+
+<div
+    class="flex h-auto w-full flex-col overflow-hidden rounded-2xl bg-csi-neutral-50 shadow-lg md:h-[32rem] md:w-72 dark:bg-csi-neutral-900"
+>
+    <img {src} alt={title} loading="lazy" class="m-0 h-64 shrink-0 object-cover md:h-48" />
+    <div class="m-3 flex h-full flex-col justify-between gap-2 overflow-hidden">
+        <p class="m-0">{type}</p>
+        <h2 class="m-0">{title}</h2>
+        <p class="m-0 grow overflow-hidden">{description}</p>
+        <p class="m-0"><a href={url}>More Info</a></p>
```
ehrelevant, line 22:
```suggestion
    export let href: string;
</script>

<div
    class="flex h-auto w-full flex-col overflow-hidden rounded-2xl bg-csi-neutral-50 shadow-lg md:h-[32rem] md:w-72 dark:bg-csi-neutral-900"
>
    <img {src} alt={title} loading="lazy" class="m-0 h-64 shrink-0 object-cover md:h-48" />
    <div class="m-3 flex h-full flex-col justify-between gap-2 overflow-hidden">
        <p class="m-0">{type}</p>
        <h2 class="m-0">{title}</h2>
        <p class="m-0 grow overflow-hidden">{description}</p>
        <p class="m-0"><a {href}>More Info</a></p>
```
Same changes for the respective Event/Project type definitions and Event/Project objects in `events.ts` and `projects.ts`. Not sure why I used `url` here back then.

src/lib/components/EventCard.svelte
```diff
@@ -1,18 +1,29 @@
...
     // eslint-disable-next-line init-declarations
-    export let alt: string;
+    export let state: string;
+    // eslint-disable-next-line init-declarations
+    export let title: string;
+    // eslint-disable-next-line init-declarations
+    export let type: string;
     // eslint-disable-next-line init-declarations
-    export let href: string;
+    export let url: string;
```
ehrelevant, line 15:
```suggestion
    export let href: string;
```

src/lib/components/EventCard.svelte
```diff
@@ -1,18 +1,29 @@
...
-    <img {src} {alt} loading="lazy" class="m-0 h-64 shrink-0 object-cover md:h-48" />
+    <img {src} alt={title} loading="lazy" class="m-0 h-64 shrink-0 object-cover md:h-48" />
     <div class="m-3 flex h-full flex-col justify-between gap-2 overflow-hidden">
-        <slot />
-        <p class="m-0"><a {href}>More Info</a></p>
+        <p class="m-0">{state} - {type}</p>
+        <h2 class="m-0">{title}</h2>
+        <p class="m-0">{dates}</p>
+        <p class="m-0 grow overflow-hidden">{description}</p>
+        <p class="m-0"><a href={url}>More Info</a></p>
```
ehrelevant, line 27:
```suggestion
        <p class="m-0"><a {href}>More Info</a></p>
```

src/lib/components/FilterPagePanel.svelte
```diff
@@ -1,37 +1,29 @@
...
+    import type { Project } from '$lib/projects/projects';
+    import ProjectCard from './ProjectCard.svelte';
 
     // eslint-disable-next-line init-declarations
     export let filters: string[];
     // eslint-disable-next-line init-declarations
-    export let events: Event[];
+    export let cardsInfo: (Event | Project)[];
+    // eslint-disable-next-line init-declarations
+    export let cardComponent: typeof EventCard | typeof ProjectCard;
```
ehrelevant, line 12:
```suggestion

```

ehrelevant (changes requested):
First of all, I'd like to apologize over this task since while I expected it would be messy just from writing the task, it ended up being a whole lot stranger than I initially expected.

Second of all, I'd also like to apologize for (maybe) putting too many changes into a single commit and the poor commit description. That was my bad, for the latter I more or less ran out of ideas on what to name the commit since, as mentioned, it was too large.

As someone partially responsible for these changes, I'm still unsure whether attempting to generalize, especially to this degree, was the right decision. Although, I suppose this dilemma will now be left to the reviewers to decide.

Anyhow, I wrote down some other minor changes that could be done here and there to improve the implementation.

syncopascual:
Committed the suggestions from above! Also changed a few things here and there to accommodate these:

- Changed the url properties in the Event and Project interfaces and cards to href instead
- Removed the cardComponent prop in both EventCard and ProjectComponent; now the rendering of an event vs a card is decided by whether the cardsInfo passed to the panel (effectively this also meant that the event and project `+page.svelte` no longer import EventCard or ProjectCard; instead importing the components is done by FilterPagePanel)

src/lib/projects/projects.ts
```diff
@@ -1,15 +1,21 @@
-import type { Event } from '$lib/components/EventsPanel.svelte';
 import lino from '$lib/lino-sablay.svg';
 
+export interface Project {
+    tag: string;
+    type: string;
+    title: string;
```
ehrelevant, line 6:
```suggestion
    project: string;
```

src/lib/events/events.ts
```diff
@@ -9,6 +7,17 @@ import gitWorkshop from '$lib/events/git-workshop.webp';
 import pyrolympics from '$lib/events/pyrolympics.webp';
 import yearEndGa2023 from '$lib/events/year-end-GA-2023.webp';
 
+export interface Event {
+    tag: string;
+    state: string;
+    type: string;
+    title: string;
```
ehrelevant, line 14:
```suggestion
    event: string;
```

src/lib/components/FilterPagePanel.svelte
```diff
@@ -80,15 +70,12 @@
...
-                <EventCard {src} alt="image of {title}" href={url}>
-                    <svelte:fragment>
-                        <p class="m-0">{state} - {type}</p>
-                        <h2 class="m-0">{title}</h2>
-                        <p class="m-0">{dates}</p>
-                        <p class="m-0 grow overflow-hidden">{description}</p>
-                    </svelte:fragment>
-                </EventCard>
+            {#each filteredCardsInPage as cardProps}
+                {#if 'state' in cardProps}
```
ehrelevant, line 74:
```suggestion
                {#if 'event' in cardProps}
```

ehrelevant (approved):
> Also, I suppose rather than `state` and `dates`, it might be better to change `title`—in both the `Event` and `Project` interfaces—to `event` and `project`, then base it off that.

Just wanted to bring up this comment again, since I still think this will make it seem cleaner. I added in put in some suggestions to give an idea of what I mean (didn't put in everything since it'll be too cluttered). What do you think?

That aside, this LGTM! Though, it'll be up to the actual reviewers to decide on whether we should really be going in this direction for this component.

syncopascual:
Committed these changes too! I'm also kinda partial to the intuitiveness of this rather than looking for state and/or date (since we might want to use those for the project component later on if we add extra info to the cards). Thank you!

VeeIsForVanana (approved):
I think @ehrelevant handled most of the concerns on this. I'll issue my approval here and merge in a moment.

PR #92, opened 2024-10-11
feat: split member details into JSON files

### LINKED ISSUE #77: Split team member details into separate files
See discussion in #76.

ehrelevant:
I apologize in advance if this PR is a bit messy. I'm a bit busy studying for an exam right now, but I'll see if I can clean it up tomorrow evening. 

This PR closes #77. In short, this PR enables storing of member details for the People page as individual JSON files in the `$lib/people/team/members/` directory. This also enables the storing of member images in the  `$lib/people/team/images/` directory. 

This was implemented through the use of vite dynamic/glob imports to collate the files on build rather than on run. More details can be seen in the comments on the implementation. It is also due to this that JSON was preferred over using `ts`/`js` files.

Due to the way this was implemented the following requirements must be enforced for the function to collate member details and images correctly:
- Images must be stored with the `.webp` filetype ([limitation of vite/rollup dynamic imports](https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#imports-must-end-with-a-file-extension))
- The following details of each member must be included in the JSON file:
  - `name` — Name of member
  - `title` — Title of member (i.e. President, Member, etc.)
  - `committee` — Full name of the committee, refer to `$lib/people/team/team.ts` for options (must be valid, otherwise will raise an error)
  - `img` —  File name of member's `.webp` image
  - `socials` — Record/"dictionary" of a member's social media links; this should be written as `socials: {}` if not providing any social media links


src/lib/people/team/team.ts
```diff
@@ -0,0 +1,64 @@
...
+
+            /**
+             * Other issues with JSON values should raise an error without us having to explicitly account for them.
+             * Might be good to improve this later, but this suffices for the moment.
+             */
+
+            assert(
+                committees.includes(memberObj.committee),
+                `committee "${memberObj.committee}" of "${memberObj.name}" does not exist`,
+            );
```
BastiDood, line 57: We don't actually have to `assert` here if we parse the JSON using a [Zod](https://zod.dev/) or [Valibot](https://valibot.dev/) schema. This has the benefit that we don't just blindly type-cast the `import.meta.glob` with `<Member>`.

```ts
// Zod
import { Member } from '$lib/models/member';

const value = import.meta.glob('*.json');
const json = Member.parse(await value['/']);
```

```ts
// Valibot
import { Member } from '$lib/models/member';
import { parse } from 'valibot';

const value = import.meta.glob('*.json');
const json = parse(Member, await value['/']);
```

> [!NOTE]
> I personally prefer Valibot over Zod, but either is fine.

src/lib/people/team/team.ts
```diff
@@ -0,0 +1,64 @@
...
+     * The usage of dynamic imports in this function will have no effect on page load times.
+     * This is because dynamically imported assets will still be bundled on build.
+     * Thus, this will make things more organized over storing the images in "static/".
+     */
+    const teamPromise = import.meta.glob<Member>('./members/*.json', {
+        import: 'default',
+    });
+
+    const team: Member[] = [];
+    for (const path in teamPromise) {
```
BastiDood, line 32: This can be refactored into a much cleaner version with [`Promise.all`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) and [`Object.entries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries).

> [!IMPORTANT]
> In my implementation, I removed the `img` field in the `Member` JSON because we can actually derive the intended file name on the import path anyway.

```js
import { basename } from 'node:path';

const imports = import.meta.glob('$lib/members/*.json');
const promises = Object.entries(imports).map(async ([path, asset]) => {
    // Derive the image name from the original path
    const base = basename(path, '.json');
    const img = await import(`$lib/people/team/images/${base}.webp`).default;

    // See my other suggestion on Zod and Valibot for more details.
    const member = parse(Member, await asset);

    // Type-safety enforced at build-time and compile-time!
    return { ...member, img };
});

// Await all imports in parallel!
const members = await Promise.all(promises);
```

src/routes/people/+page.ts
```diff
@@ -0,0 +1,7 @@
+import getTeam from '$lib/people/team/team';
+
+export async function load() {
+    return {
+        team: await getTeam(),
+    };
```
BastiDood, line 6: This is a personal nitpick on code style; feel free to ignore if you disagree.
```suggestion
    return { team: await getTeam() };
```

BastiDood (changes requested):

src/lib/people/team/team.ts
ehrelevant: I didn't initially consider deriving the intended file name for the import path, but I really like your approach so I'll try adding it to the PR. Thanks for that and the suggested refactor!

That said, I've been having difficulty getting `node:path`, which was used in the provided code, to work as expected. Looking it up, [it doesn't seem to mix particularly well with Vite](https://vite.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility). That aside, when I tried using polyfills for it (i.e. [vite-plugin-node-polyfills](https://www.npmjs.com/package/vite-plugin-node-polyfills)), I ran into a problem where it breaks the dev server (`500 Internal Error`) upon _reloading_ the People page. While this doesn't seem to be an issue on build, it seems like it would be annoying to deal with. 

Anyhow, I'll be opting for a normal typescript or regex approach for the filename rather than using `basename` until this issue gets figured out if that's alright.

BastiDood: > That said, I've been having difficulty getting node:path, which was used in the provided code, to work as expected.

Ah, I got it. We need to rename the `+page.ts` to `+page.server.ts` so that it only runs in the dev server (i.e., a Node.js environment) and the build step (i.e., a Node.js environment). The `+page.ts` tells SvelteKit to emit client-side bundles, which would indeed break when we import `node:path`. But if we use `+page.server.ts` instead, the `adapter-static` will emit `.json` files in the `build` folder.

## Resources
* [The Missing Guide to Understanding `adapter-static` in SvelteKit](https://khromov.se/the-missing-guide-to-understanding-adapter-static-in-sveltekit/)

src/lib/people/team/team.ts
```diff
@@ -0,0 +1,47 @@
...
+    'Branding & Creatives',
+    'Engineering',
+];
+
+const MemberSchema = v.object({
+    name: v.string(),
+    title: v.string(),
+    committee: v.pipe(
+        v.string(),
+        v.check(i => committees.includes(i), 'invalid committee'),
```
BastiDood, line 18: Let's use [`picklist`](https://valibot.dev/api/picklist/) for this. 👍

src/lib/people/team/team.ts
```diff
@@ -0,0 +1,47 @@
...
+
+        const member = v.parse(MemberSchema, await asset());
+
+        // Type-safety enforced at build-time and run-time!
+        return { ...member, img };
+    });
+
+    const members = await Promise.all(promises);
+
+    return members;
```
BastiDood, line 46: We can inline this.
```suggestion
    return await Promise.all(promises);
```

src/lib/people/team/team.ts
```diff
@@ -0,0 +1,47 @@
...
+const baseRe = /(?<base>\w+).json$/;
+
+export async function getTeam(): Promise<Member[]> {
+    const imports = import.meta.glob<Member>('./members/*.json');
+
+    const promises = Object.entries(imports).map(async ([path, asset]) => {
+        // Derive the image name from the original path
+        // It goes without saying that this may not be the right approach for this
+        const base = path.match(baseRe)?.groups?.base;
+        const img: string = (await import(`$lib/people/team/images/${base}.webp`)).default;
```
BastiDood, line 36: Why was the manual `string` type annotation necessary here? Shouldn't TypeScript be able to infer this anyway?

src/lib/people/team/team.ts
```diff
@@ -0,0 +1,47 @@
...
+    socials: v.optional(v.record(v.string(), v.pipe(v.string(), v.url()))),
+});
+
+export interface Member extends v.InferOutput<typeof MemberSchema> {
+    img: string;
+}
+
+const baseRe = /(?<base>\w+).json$/;
+
+export async function getTeam(): Promise<Member[]> {
```
BastiDood, line 29: We can actually omit the return type here because TypeScript is smart enough to infer this from the function's `return` value.
```suggestion
export async function getTeam() {
```

src/routes/people/+page.svelte
```diff
@@ -1,4 +1,4 @@
-<script>
+<script lang="ts">
```
BastiDood, line 1: No need for TypeScript here. As far as I can tell, the syntax here is all sufficiently JavaScript.
```suggestion
<script>
```

src/routes/people/+page.ts
BastiDood, line 1: See https://github.com/up-csi/up-csi.github.io/pull/92#discussion_r1798740946. Moreover, we can actually just use `+page.server.js` since there is no TypeScript-specific syntax used here.

BastiDood (changes requested):

src/lib/people/team/team.ts
```diff
@@ -0,0 +1,47 @@
...
+    committee: v.pipe(
+        v.string(),
+        v.check(i => committees.includes(i), 'invalid committee'),
+    ),
+    socials: v.optional(v.record(v.string(), v.pipe(v.string(), v.url()))),
+});
+
+export interface Member extends v.InferOutput<typeof MemberSchema> {
+    img: string;
+}
```
BastiDood, line 25: One cool thing about TypeScript is that `type` declarations and `const` declarations exist in separate namespaces. We can actually get away with something like this.
```ts
export const Member = v.object({ ... });
export type Member = v.InferInput<typeof Member>;
```

src/lib/people/team/team.ts
```diff
@@ -0,0 +1,47 @@
...
+    'Executive',
+    'Innovation',
+    'Service',
+    'External Relations',
+    'Membership & Internals',
+    'Branding & Creatives',
+    'Engineering',
+];
+
+const MemberSchema = v.object({
```
BastiDood, line 13: Typically, I put these models inside a `$lib/models` or `$lib/server/models` module to keep the data model separate from the app logic. Personally, I prefer putting it in `$lib/server/models` by default and then moving to `$lib/models` when the browser bundle requires it.

BastiDood (changes requested):

src/lib/people/team/team.ts
ehrelevant: Ah thanks, this was exactly what I was looking for!

src/lib/people/team/team.ts
BastiDood, line 36: We can probably do something cute like this just to avoid the weird-looking parentheses.
```suggestion
        const { default as img } = await import(`$lib/people/team/images/${base}.webp`);
```

ehrelevant: It actually doesn't infer it, since default evaluates to `any`. This is probably because the return type of `default` (and import for that matter) depends on the file that is eventually imported (i.e. JSON -> Array/Object, image -> string).

BastiDood: Oh, that's funny. I think I know why that happens. The TypeScript LSP figures out the return type for `import` based on the `declare module` declarations in the app (including `node_modules`). Vite specifically adds a `declare module` for `*?url` (i.e., all modules that end in `?url`), which allows TypeScript to infer the import as `string`. Since we only care about the URLs anyway, let's use the `?url` import instead of the usual bare path. If even that doesn't work, I have a few more tricks up my sleeve that aren't as clean. 😅

src/lib/people/team/team.ts
ehrelevant: Hmm, I'll take note of being able to use `Member` for defining the schema, but I think I'll prefer `interface`s for the `Member ` type declaration since it needs to include `img` as well. Extending an interface seems more appropriate than extending a type. (Also, I realize we could rename `img` to `src` at this point)

BastiDood: Yup, I'm on board with that. The `type` was used as an example, but I believe `interface` declarations are also in a separate namespace from variable names. It should work in the same way (if I recall correctly).

VeeIsForVanana:
Apologies for the late inform, but I'd like to know if we could add a field to the member JSON so we can keep track of what academic year the JSONs are for. This way it'd be easier to perform cleaning on the JSONs, and so that we could do some automated detection on the frontend to prune out all former members.

ehrelevant:
> This way it'd be easier to perform cleaning on the JSONs, and so that we could do some automated detection on the frontend to prune out all former members.

Hmm, I feel like this wouldn't be very useful since I imagine we'll have to clear the JSONs every semester prior to accepting PRs for the new semester. In that case, wouldn't it suffice to just clear it out manually rather than automatically (especially considering that committee assignments don't release at set times)? Though, if this is really necessary then would it be fine if its format was like `2425a`/`2425b`?

src/lib/people/team/team.ts
ehrelevant: Doing some testing, `*?url` really doesn't seem to have any effect on type inferences for vite dynamic imports, but it does work for normal static imports. Can I hear some of those other ideas?

src/routes/people/+page.ts
ehrelevant: Comparing `+page.server.js` and `+page.server.ts`, using the former causes the type of load `data` in `+page.svelte` to infer `any` for me. So, I think I'll opt to keep it as `+page.server.ts` rather than manually setting types.

src/lib/people/team/team.ts
ehrelevant: For now, I won't be changing this part (besides adding `?url` to the import) or applying your suggestions here due to the aforementioned [type inferencing issues mentioned in a previous reply](https://github.com/up-csi/up-csi.github.io/pull/92#discussion_r1799220272).

src/lib/models/member.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const committees = [
+    'Executive',
+    'Innovation',
+    'Service',
+    'External Relations',
+    'Membership & Internals',
+    'Branding & Creatives',
+    'Engineering',
+];
```
BastiDood, line 11: As a matter of convention, I prefer naming (true) constants in `SCREAMING_CASE` just to signify that the values are not your typical variables in `camelCase`. I also suggest adding an `as const` cast to the data so that TypeScript infers the array as [`readonly`](https://www.typescriptlang.org/docs/handbook/utility-types.html#readonlytype).

```suggestion
export const committees = [
    'Executive',
    'Innovation',
    'Service',
    'External Relations',
    'Membership & Internals',
    'Branding & Creatives',
    'Engineering',
] as const;
```

VeeIsForVanana:
> Hmm, I feel like this wouldn't be very useful since I imagine we'll have to clear the JSONs every semester prior to accepting PRs for the new semester. 

Ah, fair point. Let's drop that suggestion.

BastiDood (approved):
Love this, great work! LGTM! 🚀

ehrelevant:
oops premature push, give me a sec.

BastiDood:
> oops premature push, give me a sec.

No bro, it's also a premature approval on my end. 😅

VeeIsForVanana (approved):
If everything's all good, here's the all clear. Will merge in a bit. I'll next send off the announcement to the exec and the rest of the membership to start merging these in. Is there anything that we should highlight for them when they make their requests other than standardizing formats and conventions @ehrelevant ?

ehrelevant:
We should probably tell them to make sure that `pnpm build` runs properly before pushing (to easily validate the schema). Also, mention that the JSON filename has to match the image filename. 

That aside, I just realized we should remove the Calebs and replace it with a serious entry which they can reference before making the announcement. (Would you like to do the honors @VeeIsForVanana?)

VeeIsForVanana:
Copy. Actually what I'm thinking of is, instead of integrating it into this PR, I'll send off my own PR in the appropriate format and have them reference that instead. For the meantime, let's just get rid of the placeholder Calebs.

BastiDood:
> Actually what I'm thinking of is, instead of integrating it into this PR, I'll send off my own PR in the appropriate format and have them reference that instead.

Just to add from my experience, we typically off-shoot these kinds of additional work (tangential to the current PR) as a follow-up PR. If the follow-up PR cannot be worked on immediately, it's worth adding that as a high-priority ticket. This is mostly in an effort to sync up with `main` as soon as possible so that the other team members can work on the latest version.

> For the meantime, let's just get rid of the placeholder Calebs.

👀

VeeIsForVanana:
Okay wait let's keep the placeholder Calebs just so we can merge this in ASAP. I'm also worried we'll still need a reference for the member JSON format.

PR #93, opened 2024-10-13
fix: remove phone number from footer

### LINKED ISSUE #91: Remove phone number from contacts
Remove the placeholder phone number currently in the footer and in all contact forms, we should not be using phone numbers at all as a contact method for the organization as a whole.

BastiDood:
I initially just wanted to update dependencies, but (in the spirit of Hacktoberfest) I might as well close #91 while I was already there.


VeeIsForVanana:
LGTM. Merging in a few moments.

VeeIsForVanana (approved):

PR #94, opened 2024-10-15
fix(TeamCard): fix issue with member images not expanding correctly

ehrelevant:
This PR adds a missing `w-full` that's supposed to expand the member images on the People page correctly. In the future, we might want to consider using [`enhanced:img`](https://kit.svelte.dev/docs/images#sveltejs-enhanced-img) instead, but this should do for an urgent change.


ehrelevant:
Oh, I just realized, that `md:h-auto` should also be `md:h-full` I think.

PR #97, opened 2024-10-22
refactor!: migrate to Svelte 5 🎉

BastiDood:
Svelte 5 was finally released over the weekend. The [new documentation site](https://svelte-omnisite.vercel.app/) is still pending publication, but I've already went ahead and applied the mechanical/syntactic changes to the code. Here is a summary of the changes:

* Migrated to `vite-plugin-tailwind-purgecss` for all Tailwind purging needs.
* Disabled the aggressive minification of HTML comments because SvelteKit depends on the generated comments for efficient hydration.
* Moved some `devDependencies` that should be in `dependencies`.
* Migrated from `@steeze-ui` to `@iconfiy` icons.
* Migrated to using `$state` and `$derived` for many variables.

## Future Work
The `$lib/stores/color-scheme` needs a major refactor away from the old-style stores to the new-age runes. I've left this as a to-do item in the meantime to keep this PR well scoped.


BastiDood:
## Resources
* [Announcement Blog Post](https://svelte.dev/blog/svelte-5-is-alive)
* [Migration Guide](https://svelte.dev/docs/svelte/v5-migration-guide)

> [!NOTE]
> The old `svelte.dev`, `kit.svelte.dev`, and `learn.svelte.dev` websites have now been merged into a single `svelte.dev` website under a new brand identity.

package.json
```diff
@@ -14,46 +15,44 @@
...
         "lint:svelte": "svelte-check --tsconfig ./tsconfig.json"
     },
+    "dependencies": {
+        "@iconify/icons-heroicons": "^1.2.9",
+        "@iconify/icons-simple-icons": "^1.2.74",
+        "@iconify/svelte": "^4.0.2",
+        "@sveltejs/kit": "^2.7.2",
+        "svelte": "^5.0.5",
+        "valibot": "^0.42.1"
+    },
```
ehrelevant, line 28: A bit of a nitpick, but I might as well.

Aren't these usually put in `devDependencies`? I suppose Svelte 5 and Iconify apparently both rely on run time (though in the [latter](](https://iconify.design/docs/icon-components/#process)) it seems it's only the API call that is done on run time). Also, I think `valibot` can be moved to `devDependencies` needed since it isn't used on run time.

ehrelevant (approved):
Thanks for the refactor and PR! Really appreciate that you're still contributing to the website!

I'll be approving this, though I do have some minor questions regarding the choice of moving some `devDependencies` to `dependencies`. **Otherwise, LGTM!**

As a side note, it's nice to finally see Svelte 5 released after waiting for so long. Really excited to see what all of its features will bring to this website. Reading the changes, it seems to address a lot of my own issues working with Svelte 4. In any case, I think that this switch to Svelte 5 should be beneficial for the development of the website in the long run.

package.json
BastiDood: ## Motivation

That's true, but I've recently taken a more conservative stance on `devDependencies` and bundled projects after my experience with [DRAP](https://github.com/BastiDood/drap). Recall that Vite automatically decides whether to include snippets of `node_modules` into the browser bundle.

Usually, the heuristic is simple: if we import a module in the client side (i.e., using SvelteKit's definition/boundaries), then Vite bundles it into the browser build. Otherwise, if we import a module in the server side (e.g., `*.server.js`), Vite foregoes bundling it into the server build and instead opts to externalize the module (a.k.a. delegating to the original `node_modules`).

## So where is the problem?

By moving a dependency _into_ the `devDependencies`, we force Vite to _always_ bundle and _never_ externalize the dependency, which is quite relevant to us as we only use the SvelteKit "server side" to build the pre-rendered production build statically. In other words, the distinction between externalized server modules and force-bundled server modules is not really important because we only care about being able to import the code.

Where it _does_ matter, however, is when we use `adapter-node` (or pretty much any other adapter that runs in a Node.js server context alongside a `node_modules` folder). When we force a dependency to be included in the server bundle (via `devDependencies`), its code is copied-and-pasted into the resulting `build/index.js` despite it already being present in the `node_modules`.

This duplication is frankly a waste of space that will be relevant when considering the space constraints of lambda environments. The solution, of course, is to maintain the default behavior of `dependencies` so that we can deduplicate by externalizing/delegating to `node_modules` instead.

## Seeing It in Action

We can see what I mean by this by running [`pnpm deploy`](https://pnpm.io/cli/deploy), a helpful command that bundles a project's source code (i.e., the final `build/`) and its production `dependencies` (i.e., in the `node_modules`) into a single self-contained folder for deployment. All `devDependencies` are purged. Only the `dependencies` are included in the deployed `node_modules`.

We thus have a simple rule to follow: if the server imports the module, then put it in the `dependencies` so that it's externalized when running `pnpm deploy`. But interestingly, this is also the rule for the client: if the client imports the module, then put it in the `dependencies` so that Vite bundles it into the browser build.

**In any case, then, we put client-imported and server-imported dependencies into the `dependencies`.** This is the stance that I take nowadays.

> [!NOTE]
> I'm aware that this argument is irrelevant to us right now because we use `adapter-static`, not `adapter-node`. But, I am on the camp where we should be as forward-compatible as possible. Maybe in the future, we would need to use `adapter-node` for the website (perhaps for email endpoints?). By following the correct defaults as early as now, we ensure that future maintainers already have the correct bundling strategy _by default_.

VeeIsForVanana (approved):
LGTM. I don't really have anything much to comment, most of the changes look like they're in order and the dependencies issue appears to have been resolved.

Issue #82, opened 2024-10-01
Convert executive board sections into components


ehrelevant:
As mentioned by @galierie, it would be a good idea for us to convert the "executive board sections" on the People page into components. Additionally, we should implement it similarly to `team.ts` or `events.ts`, where we store the "JSON" information of the executive boards in a `.ts` file and get the data from there. It's also worth noting everything mentioned in #76 when implementing this, including avoiding directly doing a `default export` and having multiple files for storing the "JSON" data for each board. 

I'll likely go back to edit this issue description once the next set of tasks has been formalized. In the meantime, this issue will be assigned to @galierie!

ehrelevant:
Completed in #98.

PR #98, opened 2024-11-07
feat: make component for exec board display

galierie:
This PR addresses #82. Summarizing the changes:

- Implement something similar to `team.ts` instead of `events.ts` to store and get the executive board details. Seems a bit overkill at the present but I realized that a single `ts` file holding all executive board details might be too crowded na by next year.
- Create a dedicated component to display the executive board details (`ExecPanel`).
- Make schemas and interfaces for the executive boards (`Board`) and their members (`Officer`). 
- Rename some files


ehrelevant (approved):
Looked through the changes, was admittedly a bit caught off-guard by the `try...catch` but I get why that has to be there unlike in `team.ts`. Not gonna lie, this implementation feels very clean, though yeah it does seem a bit overkill right now 😅. Anyways, LGTM!

VeeIsForVanana (approved):
I don't think I can contribute much else to this conversation. This looks pretty nice!

LGTM! 🚀 🚀 🚀

VeeIsForVanana:
I'll request pictures of the ExeComm sometime soon so we can fill these out. Let's address that in a separate PR.

PR #99, opened 2024-11-12
feat: update events

Ulyzses:
This PR updates the events found in the `/events/` page of the site as per [this doc](https://docs.google.com/document/d/1iVgtlf-cMMEhTb1bJrKmpsK0c8-fN76zhUh8DSdhoDw/edit?tab=t.0).

Some points of conversation related to this contribution:

1. Should we keep the past events that were overwritten with the new ones. Examples of this are CSFlight, Bootcamp 10.0, Git Workshop, GAs.
2. There were some items that overlap e.g. Application Process and Pyrolympics (which I experienced as part of my App Process). Does that necessitate two event posts or just one (i.e. Application Process and then Pyrolympics within the specific event page).
3. The pictures provided are not `webp` files and look very scuffed and blurred. Might be better to have dedicated pubs for this instead of event documentation pics.
4. Was it within the scope of my task to fix the featured event?

Action items for future contributions

- [ ] Remove unused photos (to be done once other changes are approved)
- [ ] Change to the correct date for all events
- [ ] Create the specific event page for each event
- [ ] Update the Featured Event banner


src/lib/events/events.ts
```diff
@@ -44,86 +44,88 @@ export default [
...
-        type: Type.Internal,
-        event: 'CSFlight',
+        state: State.Present,
+        type: Type.External,
+        event: 'Application Process',
         dates: 'August 4 - August 6, 2023',
         description:
-            'Qui eu voluptate non nostrud. Et irure reprehenderit pariatur nulla laboris ipsum ut tempor magna officia tempor mollit Lorem reprehenderit.',
-        src: csflight,
+            'The application process serves as the recruitment phase of UP CSI. During this process, applicants undergo necessary training and development to prepare them for membership in the organization.',
```
ehrelevant, line 72: I don't think most of these description texts don't fit in the cards, so it might be good to adjust either the card sizing or text sizing slightly.

You don't need to handle this since it's not really part of your task right now, we can put this up as a separate minor issue for whoever wants to contribute. Still, it should be mentioned as a problem.

ehrelevant (changes requested):
Thank you for the PR! going over your points:

> 1. Should we keep the past events that were overwritten with the new ones. Examples of this are CSFlight, Bootcamp 10.0, Git Workshop, GAs.

Given the lack of descriptions at the moment, I think we should remove them until we can obtain them from other committees. My bad though, I should’ve checked the gdoc more thoroughly when Victor sent it to me. 

> 2. There were some items that overlap e.g. Application Process and Pyrolympics (which I experienced as part of my App Process). Does that necessitate two event posts or just one (i.e. Application Process and then Pyrolympics within the specific event page).

For cases like Pyrolympics and Application Process, I think it's fine to put these separately. However, that’s only if we have the details/descriptions to distinguish them. Otherwise, removing whatever events do not currently have details on the doc would be better.

I suppose another question here is how specific events on the Events page should be. As you mentioned, should we have an all-encompassing event card for “Application Process” or each individual “Application Process” (i.e. Elevate, Ignite) we’ve done over the years as a card? Personally, I’d say the latter seems more appropriate. However, as it stands, we’ll need to request more details from M&I and externals before we can do that.

> 3. The pictures provided are not `webp` files and look very scuffed and blurred. Might be better to have dedicated pubs for this instead of event documentation pics.

Honestly, I don't think it is a problem that they are scuffed and blurred. The images on the event/project cards are being used as thumbnails anyway. Most people won’t zoom in and notice the blurriness. Rather, I'd say it's more important that the image loads fast than be clear. That said, the pictures should indeed be `.webp` since, if I'm not mistaken, they load faster and have better compression. So, would it be possible for you to convert them first?

> 4. Was it within the scope of my task to fix the featured event?

Nope, but feel free to do so if you would like to! (Though, we really should either clean up or temporarily remove that featured event card since it doesn't serve much purpose at the moment)

>  Create the specific event page for each event

I think it would be better to use a [modal](https://flowbite.com/docs/components/modal/) with slotted (snippet-ed?) content instead of a full-on page for each event. There are plans to create such a component soon.

Besides that, put a few comments here and there and some changes that need to be made before this gets pushed (especially related to the image sizing). Pinging @VeeIsForVanana too, it’ll be better if you look through the points as well, especially points 1 and 2.

VeeIsForVanana:
Hello! Apologies for the late response. I'd just like to bump this PR, @Ulyzses please comply with the requested changes from @ehrelevant so we can move forward. In any case, I'll also be commenting on the higher-level points raised in this PR.

> 1. Should we keep the past events that were overwritten with the new ones. Examples of this are CSFlight, Bootcamp 10.0, Git Workshop, GAs.

I agree with Ehren's decisions on this. Let's remove anything that's too incomplete and looks placeholder-y while we request them from the other committees. Though I will need to know which ones these are exactly so I can issue the requests.

> 2. There were some items that overlap e.g. Application Process and Pyrolympics (which I experienced as part of my App Process). Does that necessitate two event posts or just one (i.e. Application Process and then Pyrolympics within the specific event page).

Again, I'll side with the standing decision. If the events are reasonably distinct from each other, let's keep them separate. If on the source material they're treated separately and have separate descriptions, I think it'd be fine to follow it. 

Really, generally, unless there are some gaping holes, defer to the source document so our main responsibility stays at handling content management.

At the higher level, I guess I also favor having one "event" for each application process cycle.

> 3. The pictures provided are not webp files and look very scuffed and blurred. Might be better to have dedicated pubs for this instead of event documentation pics.

I think it would be a good idea to get the original pubs for them, though the blurry event documentation pictures are fine enough if there's nothing else.

> 4. Was it within the scope of my task to fix the featured event?

I'll let Ehren's word on this stand.

Thanks and let's get this done!

galierie:
I'll be closing this na as we'll be exporting event information instead from Google Sheets. Thanks!

PR #100, opened 2024-11-25
feat(about-page): generalize button component and clean up buttons

LantingKurt:
Generalized the Button component to use <a> tags and applied the changes across the site. Updated the buttons in the About page to link properly, setting the href to # for Event and Project pages for now.


src/lib/components/LinkButton.svelte
BastiDood, line 1: I'd like to suggest extending the `HTMLAnchorAttributes` from `svelte/elements` here so that we can spread the `<a>` props to the internal one.

```svelte
<script lang="ts>
    import type { HTMLAnchorAttributes } from 'svelte/elements';
    import type { Snippet } from 'svelte';

    interface Props extends HTMLAnchorAttributes {
        children?: Snippet;
    }

    const { children, ...rest }: Props = $props;
</script>

<a {...rest} class="...">{@render children?.()}</a>
```

BastiDood (changes requested):
Just a few nitpicks here that I wanna note since the new Svelte 5 run syntax introduces nicer patterns.

src/lib/components/LinkButton.svelte
LantingKurt: Noted! Thank you for the suggestion :)

VeeIsForVanana (approved):
LGTM! 🚀🚀🚀

Issue #101, opened 2024-11-29
Add hover effects and transitions to interactive elements


ehrelevant:
I should've put this issue up earlier, oops. 

Anyhow, with the changes to the NavBar a while back, I believe it's time we start implementing `hover` effects and transitions to the interactive elements of the website (i.e. buttons, links, etc). These changes will go a long way in improving the look and feel of the website, making it more in line with modern expectations of a website (for better or worse). 

> [!NOTE]
> When implementing this, be conservative with transitions delays and durations; don't sacrifice usability and responsiveness for style. If a new website visitor notices and thinks about it, your delay/duration is likely too long. Personally, I think the default transitions offered by Tailwind have fairly sane values, so better to use those unless you have something specific in mind.

For reference, here are the pages for TailwindCSS’s [`hover:`](https://tailwindcss.com/docs/hover-focus-and-other-states) and [`transition`](https://tailwindcss.com/docs/transition-property). You may also want to look at the [documentation for Svelte transitions](https://svelte.dev/docs/svelte/svelte-transition) (alternatively, it's [tutorial page](https://svelte.dev/tutorial/svelte/transition)), but I'd advise against relying on Svelte transitions if it can be done with regular CSS/Tailwind without much hassle.

PR #103, opened 2024-12-01
feat: add hover effects and transitions

syncopascual:
This PR adds hover effects and transitions to individual components (buttons, cards, containers), as well as on the body tag in app.html for switching between dark and light mode.

I noticed some strange performance/delay issues when applying this transition Tailwind class to the EventCard and ProjectCard specifically, so I applied it to the card's parent components instead. There are probably a lot of ways to lessen the amount of places the transition tags are applied too, but I'm not sure how the inheritance works for it? I had to apply it individually on specific components in a lot of places because some text tags here and there wouldn't transition along with the parents, so it works for now.

Also, apparently there is an issue right now with how Tailwind deals with applying both linear gradients and transitions, according to [this discussion here](https://github.com/tailwindlabs/tailwindcss/discussions/4145), and the solution seems to be fading in and out of two separate components using opacity. Opted not to add the transition on the home page Hero component's gradient for now to send in the PR for the other transitions first.


src/routes/people/TeamPanel.svelte
```diff
@@ -81,7 +83,7 @@
                             : 'bg-csi-white text-csi-black dark:bg-blue-white'}
                     <li class="m-0 p-0">
                         <button
-                            class="min-h-10 shrink-0 rounded-md px-6 shadow-md {neutral} transition-colors"
+                            class="min-h-10 shrink-0 rounded-md px-6 shadow-md transition-colors duration-300 hover:bg-csi-blue dark:hover:bg-csi-black dark:hover:text-csi-white {neutral}"
```
ehrelevant, line 86: Colors nitpick, but personally I don't think `dark:hover:bg-csi-black` fits here. I think a more muted lighter color like `dark:hover:bg-white` would look nicer.

```suggestion
                            class="min-h-10 shrink-0 rounded-md px-6 shadow-md transition-colors duration-300 hover:bg-csi-blue dark:hover:bg-csi-white {neutral}"
```

src/lib/components/Button.svelte
```diff
@@ -9,7 +9,7 @@
 </script>
 
 <button
-    class="rounded-full bg-black px-4 py-3 text-center font-inter text-csi-white dark:bg-csi-blue dark:text-csi-black"
+    class="rounded-full bg-black px-4 py-3 text-center font-inter text-csi-white transition-colors duration-300 ease-in-out hover:bg-csi-blue dark:bg-csi-blue dark:text-csi-black dark:hover:bg-csi-white"
```
ehrelevant, line 12: I mentioned this in #101 (which admittedly came very late), but I'm of the opinion that hover transitions are best when they are barely noticeable. 

With that said, I believe a transition duration of 300ms for a button tends to be bordering on too much. I personally think the default duration of [TailwindCSS transitions (150ms)](https://tailwindcss.com/docs/transition-property) are fairly sane and often better to keep unless you are aiming to do something specific. This comment is also directed to every component with `duration-300`.

ehrelevant (changes requested):
Thanks for the PR! I just have a few opinions regarding the chosen transition duration and some hover colors and I would like to hear your thoughts on these.

src/routes/people/TeamPanel.svelte
syncopascual: Gets! I wasn't sure about the colors too; I'll commit the changes along with changes to the durations.

src/lib/components/Button.svelte
syncopascual: Shortened all the durations! Though I noticed that the NavBar durations that were added a bit ago are also around 200-300ms; kept those the same for now but I can shorten those too if you think that looks better as well.

syncopascual:
I think this worked? I tested it locally and the button transitions look okay, though there might be some discrepancies with the transition durations and colors that I didn't notice.

ehrelevant (commented):
Yep, the merge seems to have gone well. LGTM!!

ehrelevant (approved):

Issue #102, opened 2024-11-29
Create components for the projects page


ehrelevant:
This issue is related to #23. A few weeks ago, we received a [revised design](https://www.figma.com/design/nG8W0j19Al9LDgdscOqCoc/CSI?node-id=117-1158&t=x0LBjviwQw5dYaGk-1) for the Projects page of the UP CSI website (courtesy of Elle). I believe that the design is polished enough that there should be no issues with us making progress on this page already. 

# Tasks

Let's get this process started by preparing some of the components that we will need for this page. These components are listed below:

- [x] Social Media Panel (#107)
  - It'll contain the same social media links as the ones on the footer, but will be displaying social media icons instead of just text.
  - Take note of our usage of [Iconify](https://iconify.design/docs/icons/) for loading icons from online.

- [x] Details/Dropdown Component (#104)
  - Consider checking #79 and #98 since they are similar.
  - For the most part, this is just styling the existing component, but feel free to improve on it if you would like to (i.e. by adding animations).

- [x] General Contact Information (#106)
  - For content, just include the email on the footer and our location (Department of Computer Science, UP Diliman, etc.).

- [x] General FAQs (#105)
  - Pretty much implement it as it is.
  - We're still getting content for this, so just use Lorem Ipsum in the meantime.

Again, **the Figma design for this page can be viewed [here](https://www.figma.com/design/nG8W0j19Al9LDgdscOqCoc/CSI?node-id=117-1158&t=x0LBjviwQw5dYaGk-1)**.

~~For more details, please check the corresponding sub-issues that I'll be attaching in a moment~~ (I can't make sub-issues apparently).

# Committee Internship & Getting Started

Since it's currently committee internship season, we'll be assigning the creation of these components to our Website Team interns. I'd imagine that some of you might not be familiar with GitHub or how to get started. So, I'll just be giving a quick tutorial on how to set things up (I'll be skipping over some things).

## Setup

To start, ensure that you have installed [`Node.js`](https://nodejs.org/en) and [`pnpm`](https://pnpm.io/) (See installation guide [here](https://pnpm.io/installation)). Now, make sure to fork this repository by clicking the **fork** button (only copy `main`) and create a new fork.
![forks](https://github.com/user-attachments/assets/ee8e307e-e5f2-4fd4-95d4-3fb59749b36d)

Now, you should be on the forked repository. From here, clone the fork onto your local machine.

```bash
# Via HTTPS
[...]

PR #104, opened 2024-12-11
feat(Accordion): add dropdown component to Projects page

s0mme:
This PR addresses #102.

Summary of changes:
- Create dropdown component for details / FAQ  (`Accordion`)
- Add FAQs section to Projects page with example dropdown

Points for improvement:
- [ ] Match styling to [design document](https://www.figma.com/design/nG8W0j19Al9LDgdscOqCoc/CSI?node-id=117-1158&t=x0LBjviwQw5dYaGk-1)
- [ ] Add option for dropdowns that start out as opened (`open` property)
- [ ] Add option to freeze dropdowns (`interactive` property)

### DIFF
--- a/src/routes/projects/+page.svelte
+++ b/src/routes/projects/+page.svelte
@@ -1,4 +1,5 @@
 <script>
+    import Accordion from '$lib/components/Accordion.svelte';
     import FilterPagePanel from '$lib/components/FilterPagePanel.svelte';
     import GeneralInfo from './GeneralInfo.svelte';
     import SocialMediaPanel from '$lib/components/SocialMediaPanel.svelte';
@@ -16,3 +17,10 @@
         <SocialMediaPanel />
     </div>
 </article>
+
+<section>
+    <h1 class="mb-4 w-2/5 text-3xl text-csi-black md:text-4xl dark:text-csi-white">FAQs</h1>
+    <Accordion title="Lorem ipsum dolor sit amet?">
+        <span> Lorem ipsum dolor sit amet adipiscing elit consectitur </span>
+    </Accordion>
+</section>
--- a/src/routes/projects/+page.svelte
+++ b/src/routes/projects/+page.svelte
@@ -20,7 +20,9 @@
 
 <section>
     <h1 class="mb-4 w-2/5 text-3xl text-csi-black md:text-4xl dark:text-csi-white">FAQs</h1>
-    <Accordion title="Lorem ipsum dolor sit amet?">
-        <span> Lorem ipsum dolor sit amet adipiscing elit consectitur </span>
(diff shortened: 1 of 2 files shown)


ehrelevant (changes requested):
Thank you and congratulations on your first PR to this repository! The component you made looks beautiful and I like the implementation you did, especially your usage of ARIA attributes/states! Would just like to request a small change regarding the centering of the icons and I think it should be good to go! 

Besides that, just to address some of the points for improvement you listed:

> Match styling to [design document](https://www.figma.com/design/nG8W0j19Al9LDgdscOqCoc/CSI?node-id=117-1158&t=x0LBjviwQw5dYaGk-1)

I think your implementation looks good as-is. However, if you would like to change the component some more to fit the design document, then you could opt to do this in a later PR if you are busy right now. 

> Add option for dropdowns that start out as opened (open property)

I suppose one way this could be achieved is by passing down an `open` prop which I think you already figured out. I won't lie though, it looked a bit cursed when I tried implementing it (maybe someone more Svelte 5 experience could think of something better).

```ts
interface Props {
    title: string;
    open?: boolean;
    children?: Snippet;
}

const { children, title, open: openInit = false }: Props = $props();
let open = $state(openInit);

function toggle() {
    open = !open;
}
```
```ts
<Accordion open title="Lorem ipsum dolor sit amet?">
```

> Add option to freeze dropdowns (interactive property)

Personally, I don't think this would be necessary. There isn't any page where I can think of this being useful for us. If you do have an idea, feel free to mention it though, it would be greatly appreciated.

src/routes/projects/+page.svelte
```diff
@@ -7,3 +8,10 @@
 <article class="prose max-w-full dark:prose-invert">
     <FilterPagePanel {filters} cardsInfo={projects} />
 </article>
+
+<section>
+    <h1 class="mb-4 w-2/5 text-3xl text-csi-black md:text-4xl dark:text-csi-white">FAQs</h1>
+    <Accordion title="Lorem ipsum dolor sit amet?">
+        <span> Lorem ipsum dolor sit amet adipiscing elit consectitur </span>
+    </Accordion>
```
ehrelevant, line 16:
```suggestion
    <div class="flex flex-col gap-6">
        <Accordion title="Lorem ipsum dolor sit amet?">
            <span> Lorem ipsum dolor sit amet adipiscing elit consectitur </span>
        </Accordion>
    </div>
```
Considering there will be an accordion for each FAQ, might be worth putting it inside a `<div>` in advance.

VeeIsForVictor (approved):
I resolved the pending requested changes myself so we could get this in now. LGTM!

PR #106, opened 2024-12-19
feat: add GeneralInfo component

Nodeil:
I made a general info tile for the projects page and changed the layout to look a bit more like the wireframe. I think the responsive CSS works but I'm not really sure.

### DIFF
--- a/src/routes/projects/GeneralInfo.svelte
+++ b/src/routes/projects/GeneralInfo.svelte
@@ -1,5 +1,5 @@
 <div
-    class="m-0 rounded-2xl bg-csi-neutral-50 px-8 py-8 dark:prose-invert shadow-lg lg:max-w-fit dark:bg-csi-neutral-900"
+    class="m-0 rounded-2xl bg-csi-neutral-50 px-8 py-8 shadow-lg dark:prose-invert lg:max-w-fit dark:bg-csi-neutral-900"
 >
     <div class="flex flex-col gap-5">
         <div>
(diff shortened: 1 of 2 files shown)


src/routes/projects/GeneralInfo.svelte
```diff
@@ -0,0 +1,18 @@
+<div
+    class="m-0 rounded-2xl bg-csi-neutral-50 px-8 py-8 font-inter text-csi-black shadow-lg lg:max-w-fit dark:bg-csi-neutral-900"
+>
```
ehrelevant, line 3: The text color currently isn't changing in dark mode. One way to address this is by adding a `dark:text-csi-white` would have solved the issue.

However, instead, I suggest using `prose` and `dark:prose-invert` (from [`tailwindcss/typography`](https://github.com/tailwindlabs/tailwindcss-typography)) which applies not only the appropriate text colors but also the default font family for the website.  

```suggestion
<div
    class="prose m-0 rounded-2xl bg-csi-neutral-50 px-8 py-8 shadow-lg dark:prose-invert lg:max-w-fit dark:bg-csi-neutral-900"
>
```

ehrelevant (changes requested):
Thanks and congrats for sending your first contribution to the UP CSI website! No worries about the responsive CSS, it works great! Just commented a few changes I'd like to request before we can push these changes, so be sure to check them out as soon as you're available. Nonetheless, good work on your implementation!

VeeIsForVictor (approved):
Approved, LGTM!

PR #108, opened 2024-12-26
feat: modify card used for projects page

rue-22:
This PR modifies the implementation for the card component used in the `projects` route (i.e., `src/lib/components/ProjectCard.svelte`). This component was based on this [Figma](https://www.figma.com/design/nG8W0j19Al9LDgdscOqCoc/CSI) design by Judelle Gaza. I've removed the `href` and the `type` attribute in `projects.ts` since it doesn't really do anything to the component (correct me if I'm wrong). I've also made the `src` attribute optional so that cards without available pictures/images are supported.

### DIFF
--- a/src/lib/components/ProjectCard.svelte
+++ b/src/lib/components/ProjectCard.svelte
@@ -4,13 +4,13 @@
 </script>
 
 <div
-    class="flex h-auto w-full flex-col overflow-hidden rounded-2xl bg-csi-neutral-50 px-1 shadow-lg md:max-w-72 md:px-2 dark:bg-csi-neutral-900"
+    class="flex h-full w-full flex-col overflow-hidden rounded-2xl bg-csi-neutral-50 px-1 shadow-lg md:max-w-72 md:px-2 dark:bg-csi-neutral-900"
 >
     {#if src}
         <img {src} alt={project} loading="lazy" class="m-0 h-64 shrink-0 object-cover md:h-48" />
     {/if}
     <div class="md:text-md m-3 flex h-full flex-col justify-between gap-2 overflow-hidden">
         <h2 class="m-0">{project}</h2>
-        <p class="m-0 grow overflow-hidden">{description}</p>
+        <p class="m-0 grow overflow-hidden text-justify">{description}</p>
     </div>
 </div>
--- a/src/lib/components/ProjectCard.svelte
+++ b/src/lib/components/ProjectCard.svelte
@@ -1,16 +1,16 @@
 <script lang="ts">
     import type { Project } from '$lib/projects/projects';
-    const { src, project, type, description, href }: Project = $props();
+    const { src, project, description }: Project = $props();
 </script>
 
 <div
-    class="flex h-auto w-full flex-col overflow-hidden rounded-2xl bg-csi-neutral-50 shadow-lg md:h-[32rem] md:w-72 dark:bg-csi-neutral-900"
+    class="flex h-auto w-full flex-col overflow-hidden rounded-2xl bg-csi-neutral-50 px-1 shadow-lg md:max-w-72 md:px-2 dark:bg-csi-neutral-900"
 >
-    <img {src} alt={project} loading="lazy" class="m-0 h-64 shrink-0 object-cover md:h-48" />
-    <div class="m-3 flex h-full flex-col justify-between gap-2 overflow-hidden">
-        <p class="m-0">{type}</p>
+    {#if src}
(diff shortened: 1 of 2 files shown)


src/lib/components/ProjectCard.svelte
```diff
@@ -1,16 +1,16 @@
 <script lang="ts">
     import type { Project } from '$lib/projects/projects';
-    const { src, project, type, description, href }: Project = $props();
+    const { src, project, description }: Project = $props();
 </script>
 
 <div
-    class="flex h-auto w-full flex-col overflow-hidden rounded-2xl bg-csi-neutral-50 shadow-lg md:h-[32rem] md:w-72 dark:bg-csi-neutral-900"
+    class="flex h-auto w-full flex-col overflow-hidden rounded-2xl bg-csi-neutral-50 px-1 shadow-lg md:max-w-72 md:px-2 dark:bg-csi-neutral-900"
```
ehrelevant, line 7: I've been thinking this through for quite a bit, but I feel like it would be better if card heights along a row are all consistent. At the moment cards with different description lengths or with images look rather awkward next to each other. As an example, see the screenshot of the Projects page below. 

Screenshot 2024-12-26 at 16-14-24 UP CSI - Center for Student Innovations

Anyhow, I believe this can be quickly resolved by changing `h-auto` to `h-full`, but I'd be keen to hear your thoughts on which looks cleaner.

```suggestion
    class="flex h-full w-full flex-col overflow-hidden rounded-2xl bg-csi-neutral-50 px-1 shadow-lg md:max-w-72 md:px-2 dark:bg-csi-neutral-900"
```

src/lib/components/ProjectCard.svelte
```diff
@@ -1,16 +1,16 @@
...
 >
-    <img {src} alt={project} loading="lazy" class="m-0 h-64 shrink-0 object-cover md:h-48" />
-    <div class="m-3 flex h-full flex-col justify-between gap-2 overflow-hidden">
-        <p class="m-0">{type}</p>
+    {#if src}
+        <img {src} alt={project} loading="lazy" class="m-0 h-64 shrink-0 object-cover md:h-48" />
+    {/if}
+    <div class="md:text-md m-3 flex h-full flex-col justify-between gap-2 overflow-hidden">
         <h2 class="m-0">{project}</h2>
         <p class="m-0 grow overflow-hidden">{description}</p>
```
ehrelevant, line 14:
```suggestion
        <p class="m-0 grow overflow-hidden text-justify">{description}</p>
```

Slight nitpick, but the design uses justified text as well. Anyhow, I think it does looks better justified. We should also reflect this change in the Event cards later on if this is incorporated into this PR.

ehrelevant (changes requested):
Thanks for the contribution and apologies again for assigning this task rather late. As usual, just a few very minor suggestions on some things here and there, so do check them out as soon as you are free. Though. for now, enjoy your break and have a happy holidays!

(p.s. might take me awhile for me to review or respond to repo stuff besides this one I'm writing right now, sorry about that)

src/lib/components/ProjectCard.svelte
rue-22: Yeah, I do think it looks kind of cleaner. However, I just decided to not use it because there's a huge blank space on cards with no pictures/images (also think of cards with short description and no pictures). I thought it would be look kind of awkward and weird design-wise but, ultimately, I think it still depends on how we format the cards with one another.

src/lib/components/ProjectCard.svelte
rue-22: Ah yeah, didn't think about that. I like justified text more in general since it looks a lot cleaner. Thanks for the suggestion! Will commit another change for this.

src/lib/components/ProjectCard.svelte
ehrelevant: Yeah, that's fair enough. Personally, between the huge blank spaces on a card or cards with inconsistent sizes, I find that the latter ends up being more awkward in the context of a grid-like panel.

ehrelevant:
LGTM! Thanks again for contributing in the middle of holiday season. Let's focus on adding in the projects panel content and details next! 🚀

ehrelevant (approved):
LGTM! Thanks again for contributing in the middle of holiday season. Let's focus on adding in the projects panel content and project details next! 🚀

VeeIsForVictor (approved):
LGTM! 🚀 🚀 🚀

PR #122, opened 2025-03-13
feat(projects): fill up FAQS

mikelReinanArcardo:
This PR addresses #113 with the following changes:

- Implemented the `FaqEntry` component to populate the general FAQs.
- Each FAQ entry is sourced from the [Service Projects FAQs Google Doc](https://docs.google.com/document/d/1ToTj_l6DlYsFGosie_07WOgalok6TSJq4DK7rWcRzH8/edit?tab=t.0#heading=h.v0p1x6e5soce).

PR #123, opened 2025-03-13
fix(ci)!: patch actions

galierie:
As seen in #122, there was an error regarding `pnpm store path`. This PR fixes the workflow error by adding a step to the `build` job to install the latest `corepack` as recommended by this post from [Vercel](https://vercel.com/guides/corepack-errors-github-actions#quick-debug-steps).


VeeIsForVictor (approved):
All clear, thanks for this fix! LGTM! :rocket: :rocket: :rocket:

Issue #114, opened 2025-03-03
Rename member images


galierie:
Outlining the tasks,
- Please rename the WEBP member images to `{Nickname}{Surname}`. Note that you **can't** rename them pala directly in the GDrive, so you'll have to download the folder and rename them pa.
- Please make a PR to the website repo and upload these images in the `$lib/people/team/images` folder.

PR #128, opened 2025-03-19
feat(people): add member images

louie-github:
This should address #114. 

This PR adds member images as submitted by the members through the 2425B Renewal Form. This includes all but two members.

WebP images were simply renamed, while most PNG and JPG images were converted via `cwebp -lossless -z 9`, unless the conversion resulted in a much larger file than the original, in which case `cwebp -q 95` was used.

Pinging @galierie for approval.


louie-github:
As an additional note, some members submitted `.webp` files which were actually just renamed PNG/JPG files, which was... _not fun_ to realize. These were similarly converted as described above.

galierie (approved):
LGTM! 🚀

PR #131, opened 2025-04-01
feat: retheme for April Fools!

galierie:
This PR rethemes the CSI website to the colors of UP CURSOR.


VeeIsForVictor (approved):
This is absolutely horrifying. Kudos on the effort. Approved! Please have a recovery plan though for when 01 April finishes. I think we could just run `git revert` though. :rocket: :rocket: :rocket: 

In the future, we could rig it up so that different assets are used depending on the date, but that seems like a lot of effort for a one-off April Fools' joke.

Ulyzses:
This is actually insane

PR #130, opened 2025-03-20
refactor(Team): rearrange file system and code

galierie:
This PR mainly
- removes the type error in the `TeamPanel` component
- brings back the personal website in member JSON files

Some refactoring changes as well:
- prefer enums in picklists for the `Member` schema
- rearrange imports for `sort-imports`
  - types/enums before other imports
  - outside imports before in-repo imports
- moves the `TeamCard` and `TeamPanel` components to the components folder.
  - renames `TeamCard` to `MemberCard`

PR #132, opened 2025-04-02
feat: revert to normal theme

galierie:
This PR reverts the website to the normal theme. No resulting diffs naman after `git diff`-ing the commit with #130 so all gucci na dapat.


VeeIsForVictor (approved):
All clear, LGTM! :rocket: :rocket: :rocket:

PR #133, opened 2025-04-08
deps: update dependencies

galierie:
This PR updates all dependencies to their latest versions. This PR also removes the following dependencies:
- `vite-plugin-tailwind-purgecss`: due to a problem with building the site as a file named `loadConfig.js` couldn't be found in the package
- `autoprefixer`: in favor of `@tailwindcss/vite`

This PR also moves `svelte` from a dependency to a devDependency instead.

PR #134, opened 2025-04-08
refactor: organize `$lib`

galierie:
This PR depends on #133.

This PR specifically:
- moves all images and icons from random folders in `$lib` to `$lib/assets`
- moves all written content from random folders in `$lib` to `$lib/data`
- puts all components in `$lib/components` and organizes them in subfolders


galierie:
> This PR depends on https://github.com/up-csi/up-csi.github.io/pull/133.

Had to remove this branch's dependency on #133.

PR #139, opened 2025-05-11
chore: migrate hooks, layout, and build config to TypeScript; add path aliases

Abbott-Dev:
This PR:
- Renames `hooks.server.js` → `hooks.server.ts` and types the `handle` hook.
- Converts `+layout.js`  → `+layout.ts` with explicit exports.
- Migrates `vite.config.js` → `vite.config.ts` and adds `resolve.alias`.
- Adds `paths` in `tsconfig.json`

Now you get full TypeScript support and clean imports via `$lib`,  `$routes`, and `$static`.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,8 +1,16 @@
+import path from 'path';
 import { defineConfig } from 'vite';
 import { enhancedImages } from '@sveltejs/enhanced-img';
 import { sveltekit } from '@sveltejs/kit/vite';
 import tailwindcss from '@tailwindcss/vite';
 
 export default defineConfig({
+    resolve: {
+        alias: {
+            $lib: path.resolve('src/lib'),
+            $routes: path.resolve('/src/routes'),
+            $static: path.resolve('static')
+        },
+    },
     plugins: [enhancedImages(), sveltekit(), tailwindcss()],
 });
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -8,7 +8,7 @@ export default defineConfig({
     resolve: {
         alias: {
             $lib: path.resolve('src/lib'),
-            $routes: path.resolve('/src/routes'),
+            $routes: path.resolve('src/routes'),
             $static: path.resolve('static')
         },
     },
--- a/src/hooks.server.ts
+++ b/src/hooks.server.ts
@@ -1,12 +1,13 @@
+import type { Handle } from '@sveltejs/kit';
 import { building } from '$app/environment';
 import { minify } from 'html-minifier';
 
-export function handle({ event, resolve }) {
+export const handle: Handle = async ({ event, resolve }) => {
     let page = '';
     return resolve(event, {
-        transformPageChunk({ html, done }) {
+        transformPageChunk({ html, done }): string | undefined {
             page += html;
-            if (done)
+            if (done) {
                 return building
                     ? minify(page, {
                           collapseWhitespace: true,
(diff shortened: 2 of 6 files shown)


vite.config.ts
```diff
@@ -1,8 +1,16 @@
+import path from 'path';
```
BastiDood, line 1: Assuming we move forward with this, let's prefer using the `node:` prefix here.
```suggestion
import { resolve } from 'node:path';
```

BastiDood (changes requested):
I'm gonna have to 👎 this PR with reservations for several reasons.

> Renames `hooks.server.js` → `hooks.server.ts` and types the handle hook.

Svelte supports [zero-effort type safety](https://svelte.dev/blog/zero-config-type-safety). Explicit types are not necessary. And quite frankly, the type annotations worsen the signal-to-noise ratio of code if the types are already well implied (as in the case for `handle` function with zero-effort type safety).

> Converts +layout.js → +layout.ts with explicit exports.

As mentioned earlier, zero-effort type safety makes explicit type annotations unnecessary. But furthermore, the reason why we default to `.js` files first over `.ts` files is because there are times when a file doesn't use TypeScript syntax to begin with—as in the case for `+layout.js`.

I've written about this phenomenon more extensively two years ago in my article ["JavaScript First, Then TypeScript"](https://dev.to/somedood/javascript-first-then-typescript-10bg). The main thesis being: use `.js` by default until `.ts`-specific syntax is necessary. And in this case, type annotations (i.e., TypeScript-specific syntax) is _not_ necessary because of zero-effort type safety.

> Migrates `vite.config.js` → `vite.config.ts` and adds `resolve.alias`.

Is this _really_ necessary? Can't we just throw these to the `$lib` so that everything that needs to be shared and centralized are located there?

> Adds paths in `tsconfig.json`

See my concerns in the previous bullet.

___

Overall, it's a no from me, but those are just my two cents. If this is the convention that you guys want to follow, then by all means agree on it and move forward.

vite.config.ts
Abbott-Dev: Thanks so much for the detailed feedback and for sharing your perspective, especially the 'JS first, then TS' philosophy – it's a well-reasoned approach and I appreciate you linking the article. I agree that Svelte's advancements in zero-effort type safety and type inference are fantastic and significantly improve the developer experience by reducing boilerplate.

My thinking behind these changes, particularly the move to .ts files and some of the explicit typing, was driven by a few considerations:

1. Consistency and Explicit Clarity for Developers:
  - For `hooks.server.ts` and `+layout.ts`: While Svelte can indeed infer these types (which is great for the compiler), explicitly stating them can act as immediate documentation for any developer looking at the code. It makes the contract of the handle hook or the shape of load function parameters/return values unambiguous at a glance. The aim was to create a consistent .ts environment where types are always visible, even if sometimes inferable by the tooling.
  - I felt that the slight increase in verbosity was a worthwhile trade-off for this explicitness and for maintaining a uniform file extension for all JS/TS logic.

2. Type Safety for Configuration (`vite.config.ts`):
  - Migrating `vite.config.js` to `vite.config.ts` was primarily to leverage TS's type checking for Vite's configuration options. This can help catch typos or incorrect config structures early on, especially as Vite's config can grow in complexity.

3. Project Organization and Scalability (Aliases & tsconfig.paths):
  - Regarding `resolve.alias` and `tsconfig.paths`: While `$lib` is excellent for shared utilities, the intention behind potentially adding other aliases (like `$routes` and `$static`, or others we might define in the future) is to establish clearer import paths for different categories of modules as the project grows. This can improve organization and make it easier to navigate the codebase if it gets larger by providing semantic shortcuts beyond a single lib directory. If these specific aliases don't seem to add value yet, that's fair, but the setup enables this pattern.
  
You've made very valid points about relying on inference where possible and the signal-to-noise ratio. My approach leaned more towards prioritizing a consistently explicit TS environment across all logic files, believing it aids in long-term maintainability and team understanding, even if it means a bit more upfront typing.

If the preference is to lean more heavily on Svelte's inference and the 'JS first' approach, I'm happy to adjust. Perhaps this PR serves as a good point to discuss and solidify our preferred conventions for TS usage within Svelte projects?

Thanks again for the thoughtful review!

Issue #138, opened 2025-05-06
Restyle project cards


galierie:
See this [Figma](https://www.figma.com/design/0M2EDT1r4UWYeg6ExsTk5D/CSI_Website_Updates_2425B?node-id=31-747&p=f&t=Stx4UhtzW49xbBl0-0) from Yuwen for the design of the `ProjectCard` component (in `$lib/components/cards`). The new card style is in the design labeled "Projects Portfolio Page".

Some other notes:
- Adjust the padding and border radius as needed na lang.
- Please make sure the size of the project cards are uniform (as in, defined yung width and height ng like `px`).
  - Magmumukha siyang magulo for now. I'll handle this na lang.
  - Please limit the shown description up to 3 lines. In Tailwind, this should be a `line-clamp-3` somewhere in the tag hierarchy.

PR #141, opened 2025-05-16
feat: restyle project cards

louie-github:
This should partially address #138. The components have been restyled to match the Figma.

However, I could not make the cards have a uniform width. It seems like this is handled somewhere higher up the component tree with a CSS Grid-styled component, but I don't want to touch anything I don't really understand.

If I have time later, I'll try to see if I can figure that out.

For now, this should serve as an initial draft. Sorry for the delay!


galierie (approved):
LGTM na siya for me! Will fix the sizing issue na lang someday. Congrats! 🚀

PR #142, opened 2025-05-16
feat: create new header in Projects page

syncopascual:
This PR adds the new ProjectHeader component as designed and removes the components that are unused in the new design.


galierie (approved):
LGTM and congrats 🚀

PR #143, opened 2025-05-16
chore: add some webp files of events

syncopascual:
This adds some WebP files of pictures from InnovCamp 2020, 2023, and 2024.


galierie (approved):
Ang galing mo Marcus! LGTM 🚀

PR #144, opened 2025-05-16
feat: add carousels and modals

mikelReinanArcardo:
This PR adds the image carousels and modals for the carousel, event and project items.

Notes:
- `eslint-disable-next-line` was used for props in modals due to contradicting errors in ESLint (one variable should be `let` and the others should be `const`, but `$props()` can only be called once)
- used lino-sablay as placeholder image


galierie (approved):
Aight, LGTM and congrats 🚀

mikelReinanArcardo:
Additional notes:
- `prettier-ignore` used on the `$props()` line to avoid being formatted into multiple lines by `fmt:fix` which would then make `eslint-disable-next-line` useless

PR #147, opened 2025-10-01
perf: lazy load member images

galierie:
This PR uses dynamic imports to lazy load the images in the People Page, resulting in a faster page load. However, images don't load _agad_.


ehrelevant:
I'll be closing this PR in lieu of #149.

PR #148, opened 2025-10-04
deps!: bump major dependencies for Hacktoberfest 2025

BastiDood:
Hello all! As part of my yearly contribution to UP CSI for [Hacktoberfest 2025](https://hacktoberfest.com/), I present this year's edition of major dependency bumps. As always, all of the updated practices hail from [my SvelteKit template](https://github.com/BastiDood/sveltekit-tailwind-template). Feel free to disagree with some of the lints, and I'd gladly remove them if necessary.

## List of Refactors
- The GitHub Actions now uses `pnpm/action-setup` and `actions/setup-node` for setting up the dependency cache. No need for us to implement our own bespoke version for this.
- Removed LintHTML and Stylelint in favor of ESLint's own HTML and CSS linters.
- Updated the `eslint.config.js` to be in line with my template. Some rules have been added; some have been removed; feel free to request changes on these.
- Bumped `pnpm` from `10.8.0` to `10.18.0`.
- Assorted linter fixes on `each` blocks without keys, navigation links lacking [`resolve`](https://svelte.dev/docs/kit/$app-paths#resolve), and general code cleanliness.
- Cached Prettier and ESLint runs have now been enabled in the project.

## Future Work
A lot of components need a rewrite. Specifically, I recommend that we scrap our handwritten components to using `shadcn-svelte` (or any of your favorite libraries!) instead. I saw a lot of black magics during my refactors. 😅


ehrelevant (changes requested):
Thank you for the PR! The updated linting rules are much appreciated, and I more or less agree with all the rules that were explicitly set. Just have a couple of nitpicks that should improve cleanliness some more. Some of my comments are also directed towards @galierie, who maintained this codebase last semester and should be more familiar with it than me.

So far as components go, I do agree we will need a rewrite eventually. I also noticed a lot of weird components and some parts that had strange typing (some of which is my fault 💀). That said, I'd prefer if we complete and fix up this version of the website with handwritten components before we attempt a rewrite, since it's almost fully functional (even if it's mostly a result of black magic).

BastiDood:
Thanks for the comments @ehrelevant. A lot of the code nitpicks are due to pre-existing code that I omitted migrating over out of an abundance of caution to the existing logic. We can have a separate ticket for the actual cleanup of these components.

ehrelevant (approved):
That's fair enough, but I agree that we should put up a ticket for that. For now, let's wait for @galierie's comments, but on my end, it LGTM!!!

ehrelevant:
Additionally, on this note:

> A lot of components need a rewrite. Specifically, I recommend that we scrap our handwritten components to using shadcn-svelte (or any of your favorite libraries!) instead. I saw a lot of black magics during my refactors. 😅

In retrospect, I think I'm not actually in favor of replacing the website's handwritten components with `shadcn-svelte` or some other UI/component library. As messy as our handwritten components can (often) be, the fact that it is built by us is important as a means of training committee members in SvelteKit and Tailwind. There is also the matter of pride in our own work, but I believe @galierie could explain that argument better.

PR #154, opened 2025-12-22
feat(data): add events

mortrpestl:
- added 13 events in the `json` folder of `data`
- added images (to be placed in VPS) in this [GDrive](https://drive.google.com/drive/u/2/folders/1JQVc5HEwhXnuQpzsk37P_1o-VF012dzW)


ehrelevant:
Thank you very much for the work you put into this PR. Apologies, though, since I'll be closing this PR in lieu of the move towards using Supabase as introduced in #158. So, with respect to that, let's transfer the work done here to a new PR.

PR #158, opened 2026-03-28
refactor: use Supabase for membership data instead of JSON

syncopascual:
Refactors the backend of the website to fetch membership and executive board data from Supabase instead of using the JSON-based implementation from previous iterations of the website.

## Notes
- installs dependency @supabase/supabase-js to use the Supabase client
- Supabase client used to fetch data in the getTeam(), getOfficer(), and getExec() functions used in the People page

## Changes/fixes in progress
- There's a small inconsistency on the database with how we store links for socials
  - GitHub handles are stored as is, without any paths included (just username)
  - Linkedin and IG profiles are stored with the path included
  - We could probably just run a query to limit to just the handle so that the parsing on the backend is more consistent
  - Relevant file: `src/lib/components/cards/MemberCardTag.svelte`
- There's some boilerplate in some of the JSON parsing and types/schemas which is unnecessary now that we're getting values directly from the db
  - Specifically with the separate parsing of the string image URLs for pictures and executive board positions (ex. `2425B:Director for Engineering`), which are now separated in the db (so no parsing is needed). For now, the implementation just matches the format of the JSON implementation but using data from the db instead.
- Environment variable setup needed


ehrelevant (commented):
Sorry for the delay in the review. I don't really have any problems regarding the changes you made to add Supabase, so as far as that's concerned, nice job! 

That said, as you may notice from the changes I made, I ended up doing a couple more refactors in the codebase to clean up the models, integrate type validation into the queries, and optimize the fetches and data processing by a slight bit. Specifically, I made the following changes:
- Added parsing and type validation for environment variables
- Revised Member-related and Exec-related valibot schemas to match the database queries
- Added a step for parsing the queried data using valibot
- Replaced most of the data processing in fetching executives with a database view (see `execs_by_term` view added to the Database)
- Replaced the abbreviated committee names with their full committee names
- Removed obsolete JSON files and attempted to remove most of the unnecessary code involved in JSON parsing for the member and exec data
- Updated the GitHub actions workflow to inject the required environment variables (I made them secrets, but this should not have any effect on the deployments either way)

Anyhow, @syncopascual could you kindly review my changes before we merge this in?

syncopascual:
The changes look good to me! I'm wondering if there's a way for us to fix the GitHub Actions `build` job?

ehrelevant:
It's an issue with the secrets not being properly injected into the workflow. It normally shouldn't be hard to fix, but I may have made a mistake with the setup somewhere. I'll go make a few more attempts later before merging this in, but if it doesn't work, I'll just bypass the checks and merge this already, then I'll try fixing this directly.

ehrelevant (approved):

ehrelevant:
Alright, so the issue was that repository environment variables don't get passed to workflows in forks. I fixed this by separating the CI workflow from the build and deploy workflow, and only letting the former run in PRs. Anyhow, I'll be merging this in. Thank you very much for your contribution!

PR #161, opened 2026-04-28
deps: update dependencies

cassssy:
- Updated dependency versions in `package.json` to ESLint 10, Vite 8, and compatible related packages so the site works with the newer major releases.
- Adjusted `eslint.config.js` to disable rules that became noisy or incompatible under ESLint 10, keeping linting aligned with the current codebase.
- Removed the unused ESLint disable in `ImageCarousel.svelte` because the rule no longer applies.
- Replaced the removed EnhancedImgAttributes type in `project.ts` and `index.ts with `string[]` since the newer package no longer exports that type.
- Implemented a version override for the `cookie` package to resolve vulnerability found during `pnpm audit`.


pnpm-workspace.yaml
```diff
@@ -1,4 +1,7 @@
 onlyBuiltDependencies:
-    - '@tailwindcss/oxide'
-    - esbuild
-    - sharp
+  - '@tailwindcss/oxide'
+  - esbuild
+  - sharp
+
+overrides:
+  cookie@<0.7.0: '>=0.7.0'
```
ehrelevant: Is there any reason for adding this to `overrides`?

ehrelevant (changes requested):
Hello, thank you so much for the PR! 

Could you kindly edit in a PR description of what was changed here and the reason for the changes? Specifically, asking about the addition of `cookies` to `overrides` in `pnpm-workspace.yaml`. Additionally, it may be better to change the PR title to something more appropriate, like `deps: update dependencies`, since I don't seem to see what makes the updates to `pnpm-workspace.yaml` notable. 

Anyways, from what I can see, this is a PR focused on updating the dependencies. In that case, you may want to run `pnpm outdated` to see the outdated packages, as our current `package.json` setup makes it so that `pnpm update` only changes the minor version, but not the major version (See [Semantic Versioning](https://semver.org/)). 

Image

As far as I know, updating most of these in `pnpm outdated` simply involves changing the versions in `package.json` (e.g., `"globals": "^16.5.0"` => `"globals": "^17.6.0"`). However, there are a couple of packages (mainly, `eslint` and `vite`) that will require some more deliberate and manual work to figure out, since migrating major versions always means dealing with breaking changes and deprecations. 

Lastly, as much as possible, please make sure that the formatting (`pnpm fmt:fix`) and linting (`pnpm lint`) checks pass before committing.

pnpm-workspace.yaml
cassssy: Hi! While syncing the dependencies, I ran `pnpm audit` to check for possible security issues and it flagged a vulnerability in the dependency tree involving versions of the `cookie` package older than `0.7.0`. Running `pnpm audit --fix` automatically added this to the workspace overrides and this makes sure that even if our other libraries request an outdated version of `cookie`, pnpm forces the workspace to resolve to the secure version.

ehrelevant (changes requested):
I'd still like to follow up on bumping the major versions of the website's dependencies, as can be seen from the output of `pnpm outdated`. In particular, please migrate the code from Eslint 9 to Eslint 10 and Vite 7 to Vite 8. As mentioned, doing this will definitely involve some tinkering to fix/replace deprecated parts of the codebase, where you will likely need these relevant pieces of migration documentation for [Eslint](https://eslint.org/docs/latest/use/migrate-to-10.0.0) and [Vite](https://vite.dev/guide/migration).

Issue #160, opened 2026-04-13
Integrate supabase for event/project data and images to supabase


ehrelevant:
Currently, event/project data is still stored as JSON files in the repository, while images for both are not being hosted anywhere. These should be moved to the Supabase instance created for this website as part of #158. The data and images gathered in #154 should also be included in this transfer. 

Additionally, these need to be properly integrated into the application. This means that the models, data fetching, and data processing should be updated to remove obsolete steps related to parsing the JSON. We should also aim to make further optimizations by making full use of the increased capabilities provided by SQL operations.

mortrpestl:
Good day! 

On tasks #3 and #4 (assigned to me):

(Task 3)
Looking at the member data (from Task 2) and cross-checking, I didn't observe any discrepancies or potentially unsafe member pictures in the storage.

(Task 4) 
I have also added (from #158) in the online Supabase instance.:
- the relevant events photos
- imported the CSV containing the events data

You can view these in `events-photos` storage and the `events` database.

For reference, the next step to take (I assume this is part of #2) for this part of the website update is the integration of these events and photos to the application (by updating the data fetching methods). I hope this issue helps with the raw data necessary for testing and connecting the events to the website.

Thank you!

PR #163, opened 2026-05-10
refactor: use Supabase for events and projects data

syncopascual:
Closes issue #160 by fetching data for events and projects from Supabase.

# Implementation Notes
- Edited models to match with the Supabase schema 
- Edited `index.ts` files for projects and events
- Supabase
  - transferred old data from JSON to Supabase
  - transferred data from PR #154 to Supabase

# Misc. Notes
- When navigating to `/people`, `/projects`, and `/events` there's a slight rendering delay that happens prolly due to the data being fetched from Supabase? Instead of navigating to the page first and then the page being populated with cards as the data is fetched, the page takes a bit to load for a split second, which might feel bad especially on a worse internet connection.


ehrelevant (approved):
LGTM!! Though yeah, we will have to investigate what can be done about the initial load times, since my guess is as good as yours at the moment. 

P.S: Don't forget to set up commit signing.

ehrelevant:
Actually, wait, you mentioned "Instead of navigating to the page first and then the page being populated with cards as the data is fetched, the page takes a bit to load for a split second..." On my end, the page load is very fast, but it takes a while for the cards to load (which is somewhat expected, albeit could be improved). Was this what you were talking about, or was it something else?

